PR #6, opened 2023-07-04
Add flaw creation

d10n:
Along with many framework-level updates required to support flaw creation


d10n:
The code in this PR was previously reviewed in depth offline, so it is ready to merge, but I'll leave it open for a bit in case of any new feedback

PR #9, opened 2023-09-14
Many updates / Flaw editing

d10n:


public/service-worker.js
```diff
@@ -0,0 +1,65 @@
+const store = {};
+
+// Convenience function; debugging service workers is a pain!
```
sfxworks, line 3: Can we add a TODO here to remove this in prod?

src/components/FlawEditFormOld.vue
```diff
@@ -0,0 +1,280 @@
+<script setup lang="ts">
```
sfxworks, line 1: Instead of adding "old" shouldn't we continue to rely on git commit tracking?

src/components/IssueEditOld.vue
```diff
@@ -0,0 +1,212 @@
+<script setup lang="ts">
```
sfxworks, line 1: Same here, another old vs amending IssueEdit directly. What's the reason for this?

src/components/Login.vue
```diff
@@ -17,39 +17,49 @@ const queryRedirect = z.object({
...
+      //     name: 'index'
+      //   });
+      // })
       .catch(e => {
         console.debug('Error logging in', e);
         error.value = [
           'Error logging in.',
           'Ensure that your system has krb5 configured.',
           'Ensure that your browser has the correct trusted URIs for Negotiate authentication.',
+          'Ensure that you have logged into Kerberos on your system.',
```
sfxworks, line 57: Should we make a separate help page related to these requirements for new members that can be accessed without logging in?

src/components/widgets/EditableDate.vue
```diff
@@ -0,0 +1,363 @@
+<script setup lang="ts">
```
sfxworks, line 1: Will have to compare this later to my changes to this file but this looks good!

src/generated-client/.gitignore
sfxworks, line 1: Can we also add these to the .dockerignore?

src/stores/BackendStore.ts
sfxworks, line 1: Should we document how we are using service workers and backend here as dev notes? I can do so at a later date.

src/components/Login.vue
sfxworks: Adding link to private kba in chat description in the future

src/generated-client/.gitignore
sfxworks: Per meeting: no need as only copied in build step

src/components/FlawEditFormOld.vue
d10n: These were left in the repo for convenience of referring to recently-old rewritten code. Now removing all the old stuff.

src/stores/BackendStore.ts
d10n: Describing configuration and how it's used is good, but I think it's better to read MDN instead of redocumenting the service worker API. I'll add more configuration documentation

d10n:
Force pushed a squashed commit to clean up the mess, ready to merge

PR #18, opened 2023-10-20
Add PillList widget

d10n:


sfxworks (approved):
LGTM (will fix buildx failed with: ERROR: invalid tag ":pr-18": invalid reference format/should not apply post push given main tag)

PR #26, opened 2023-12-22
add test logout case

RichardXuan:


features/constants.py
```diff
@@ -5,6 +5,6 @@
 TIMEOUT = '5'
 
 # The follows are some elements that will be use in the test case.
-SUBMIT_BUTTON = '//button[@type="submit"]'
+LOGIN_BUTTON = '//button[text()=" Login with System GSSAPI "]'
```
d10n, line 8: To keep the test working if the HTML or text changes, it might be nice to check if the button text just contains "Login", which will probably always be in the button.
Would something like `//button[text()[contains(., "Login")]]` do that? (I haven't tested it)

d10n (approved):

features/constants.py
RichardXuan: `//button[contains(text(), "Login")]` works. Will change it in this PR.

PR #28, opened 2023-12-26
add case to test load more flaws function

RichardXuan:


features/steps/flaw.py
```diff
@@ -1,7 +1,6 @@
 from behave import given, when, then
 from selenium.webdriver.common.by import By
 
-from features.steps.common_steps import *
```
huiwangredhat, line 4: After removing this line, is there any impact on login in this scenario?

RichardXuan: No. Because common_steps.py in the dir steps, cucumber can find it automatically. I test it, works fine.

huiwangredhat:
LGTM.

huiwangredhat (approved):

PR #34, opened 2024-01-10
Support JIRA API Key in session settings

superbuggy:
- Reworks missing API key notification interface to accommodate multiple missing API keys (JIRA and BZ)
  > 
- Provides field in settings for ephemeral storage of JIRA API key, analogous to BZ KEY field (not persisted)
- Sets request header [specified in `osidb/openapi.yml` for JIRA api key](https://github.com/RedHatProductSecurity/osidb/blob/dc48ee1597f33633fe751a06a43f47bc67ca7eaf/openapi.yml#L3440-L3450)
- Provide test coverage for `<Settings/>` and its pinia store `SettingStore`

### DIFF
--- a/src/components/ToastContainer.vue
+++ b/src/components/ToastContainer.vue
@@ -75,7 +75,7 @@ function clearAll() {
   position: v-bind(position);
   top: v-bind(top);
   /*height: calc(100vh - 100px);*/
-  pointer-events: all !important;
+  pointer-events: all !important; /* override bootstrap .toast-container default value of 'none'*/ 
   height: v-bind(height);
   overflow-x: clip;
   overflow-y: auto;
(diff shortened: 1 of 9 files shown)


src/components/ToastContainer.vue
```diff
@@ -70,10 +70,12 @@ function clearAll() {
 </template>
 
 <style scoped>
+
 .osim-toast-container {
   position: v-bind(position);
   top: v-bind(top);
   /*height: calc(100vh - 100px);*/
+  pointer-events: all !important; /* override bootstrap .toast-container default value of 'none'*/
```
d10n, line 78: I usually prefer to try increasing selector specificity to avoid !important, but because pointer-events: none is surprising, this is fine

superbuggy: Likewise. I also find it surprising though I assume there's bootstrap rationale for it, so I opted for `!important` in this case to imply the override use case is very specific at present.

PR #47, opened 2024-01-25
Fix bug where owner was not able to be updated by Flaw Edit Form

superbuggy:
`owner` was missing from the schema used by the form, but no longer.

### DIFF
--- a/src/types/zodFlaw.ts
+++ b/src/types/zodFlaw.ts
@@ -131,7 +131,7 @@ export const ZodFlawSchema = z.object({
     impact: z.nativeEnum(ImpactEnumWithBlank).nullish(),
     component: z.string().max(100).nullish(),
     title: z.string().min(1),
-    owner: z.string().min(1),
+    owner: z.string().nullish(),
     trackers: z.array(z.string()).nullish(), // read-only
     description: z.string(),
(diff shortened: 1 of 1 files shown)


src/types/zodFlaw.ts
```diff
@@ -131,6 +131,7 @@ export const ZodFlawSchema = z.object({
     impact: z.nativeEnum(ImpactEnumWithBlank).nullish(),
     component: z.string().max(100).nullish(),
     title: z.string().min(1),
+    owner: z.string().min(1),
```
d10n, line 134: There might not always be an owner, so I suggest:
```
owner: z.string().nullable(),
```

superbuggy: The API will return an empty string is there is no assignee, so maybe it makes sense to remove the min, though I think (may be mistaken) the schema is 'invoked' only on form validation and not on REST ingestion. This does hint at the subtleties of where we want the gatekeeping boundaries of runtime type-checking.

 I guess the other consideration here would be to weigh `.nullish` (`<T> | undefined | null`), `.nullable` (`<T> | null`), or `.optional` (`<T> | undefined`).

d10n (approved):

PR #73, opened 2024-03-04
Run tests during development and during production build

superbuggy:


tboyetteredhat (approved):

package.json
```diff
@@ -8,10 +8,12 @@
   "license": "MIT",
   "private": false,
   "scripts": {
-    "dev": "vite",
-    "build": "run-p type-check \"build-only {@}\" --",
+    "dev": "scripts/dev-osim.sh",
+    "dev-server": "vite",
```
JakubFrejlach, line 12: So current behavior is going be that dev runs the dev server and periodically checks the test results and dev-server servers as the former dev? It should be probably added to the README.md as well.

Also I think when you run `yarn run dev` vite no longer receives commands in terminal, however that is ok I guess, autoreload still works and you can always run just the dev-server standalone without the tests.

JakubFrejlach (approved):
Small suggestion otherwise LGTM.

package.json
JakubFrejlach: Cool, thanks! LGTM now.

d10n (approved):

PR #77, opened 2024-03-05
Prefer CVE over UUID in URL for Flaws

JakubFrejlach:
This PR:
* changes the `.gitignore` pattern for `runtime.json` to enable distinguish between instance settings
* Flaw detail URLs prefer CVE ID instead of UUID

Closes OSIDB-2018


src/views/FlawEditView.vue
```diff
@@ -25,11 +27,21 @@ watch(() => props.id, () => {
...
+    return;
+  }
+
   errorLoadingFlaw.value = false;
   getFlaw(props.id)
       .then(theFlaw => {
         flaw.value = Object.assign({}, theFlaw);
         committedFlaw.value = Object.assign({}, theFlaw);
+        let pathID = (theFlaw.cve_id || theFlaw.uuid);
+        if (props.id !== pathID) {
```
superbuggy, line 41: What edge case is this  handling?

JakubFrejlach: This condition might origin from me being inexperienced with Vue js. This condition should check whether the ID in the path should actually change or not and if not, the router pushing is not happening. How big of a difference in Vue (or probably any other frontend framework) is when you push exactly same URL on router vs not pushing at all? Does that bring some potential overhead/additional rendering/etc.?

superbuggy: I think it can be left out then--  since `flaw` is a `ref` the line `flaw.value = Object.assign({}, theFlaw);` will trigger a re-render. My understanding is that any reactive values that are bound to the template will trigger a re-render, so you wouldn't have to do a reload of the page. Good consideration/question

superbuggy: Thanks for clarifying off-thread that this was for replacing the URL, that makes sense--we may want to use `history.replaceState` here instead after the flaw is refreshed. That way we can remove the flag and not have to push an additional navigation

JakubFrejlach: @superbuggy thank you! Made code changes based on our discussion, please rereview.

C-Valen: Hey guys, I didn't catch this. 
Why an additional navigation would be required on `FlawEditView.vue` without clearing the history state?

JakubFrejlach: @C-Valen I am not sure if I understand your question, but I've initially used the router.push to push URL with CVE ID instead of UUID and performed a logic with flag which preventeded refetching the flaw, however then @superbuggy correctly pointed out that history.replaceState should be sufficient. I originally avoided history.replaceState as I read in the Vue.js docs that not all globals (like window, history, etc.) are not enabled by default and got false impression that they should be avoided, however the this apparently holds only for templates on not the script part, in the script part they are available. Hopefully this explains the problem.

C-Valen: Thanks @JakubFrejlach. So the purpose of `history.replaceState` is to prevent refetching the flaw?

C-Valen: Ok, makes sense to me @JakubFrejlach. I was not seeing that scenario where an user could access directly by url to /flaws/<uuid> in which case the url indeed should be replaced to use the CVE ID (if available).
Thank you for the clarification!

C-Valen (approved):
Looks good!

src/views/FlawEditView.vue
JakubFrejlach: Thanks!

src/components/IssueQueueItem.vue
```diff
@@ -10,6 +10,7 @@ const props = defineProps<{
 const isEmbargoed = computed(() => DateTime.fromISO(props.issue.unembargo_dt).diffNow().milliseconds > 0);
 const hasBadges = computed(() => isEmbargoed.value);
 const formattedDate = computed(() => DateTime.fromISO(props.issue.created_dt).toFormat('yyyy-MM-dd'));
+const pathID = computed(() => props.issue.cve_id || props.issue.uuid);
```
superbuggy, line 13:
```suggestion
const cveOrUuid = computed(() => props.issue.cve_id || props.issue.uuid);
```

src/components/IssueQueueItem.vue
```diff
@@ -21,7 +22,7 @@ defineEmits<{
   <tr class="osim-issue-queue-item" :class="$attrs.class">
     <td><input :checked="selected" @change="$emit('update:selected', ($event.target as HTMLInputElement).checked)" type="checkbox" class="form-check-input" aria-label="Select Issue"></td>
     <td class="osim-issue-title">
-      <RouterLink :to="{name: 'flaw-details', params: {id: issue.uuid}}">{{ issue.cve_id || issue.uuid }}</RouterLink>
+      <RouterLink :to="{name: 'flaw-details', params: {id: pathID}}">{{ pathID }}</RouterLink>
```
superbuggy, line 25:
```suggestion
      <RouterLink :to="{name: 'flaw-details', params: {id: pathID}}">{{ cveOrUuid }}</RouterLink>
```

superbuggy (commented):
Solid!

src/components/IssueQueueItem.vue
JakubFrejlach: Thanks! I am doing a rebase as well so I am going to include this change as well to keep git history more clean.

src/components/IssueQueueItem.vue
JakubFrejlach: Same as above.

tboyetteredhat (approved):
LGTM 👍

PR #83, opened 2024-03-08
🐛 Prevent overlapping modals

superbuggy:
ugh


C-Valen (approved):
Code looks fine! 
Just pushed a refactor to make templates start with scripts.

PR #91, opened 2024-03-08
✨ Show OSIM version in footer

superbuggy:


package.json
```diff
@@ -13,7 +13,7 @@
     "preview": "vite preview",
     "test:unit": "vitest",
     "test:e2e": "behave features",
-    "build-only": "vite build",
+    "build-only": "vite build && node scripts/version.js",
```
superbuggy, line 16: Since this fails during build in OpenShift, maybe run this during `dev`

```diff
-    "dev": "vite",
+    "dev": "node scripts/version.js. && vite",
```

src/stores/osimRuntime.ts
```diff
@@ -17,7 +17,8 @@ export enum OsimRuntimeStatus {
   ERROR,
 }
 
-let status = ref(OsimRuntimeStatus.INIT);
+const status = ref(OsimRuntimeStatus.INIT);
+export const versionString = ref('');
```
JakubFrejlach, line 21: Maybe I would call this variable with more unambiguous name, eg. `osimVersionString`, etc.

scripts/version-plugin.ts
```diff
@@ -0,0 +1,21 @@
+import { execSync } from 'child_process';
+import { writeFile } from 'fs';
+import path from 'path';
+
+function recordVersion() {
```
JakubFrejlach, line 5: So do I understand correctly that this will run on every reload effectively helping us to not to forget the version change? I wonder maybe we should add some Github check whether the `public/version.json` file was changed so it is always guaranteed that new PR changes the version.

JakubFrejlach: Also on second thought, suppose the scenario is:
I make some changes -> reload changes the version (records the last commit) -> I commit changes -> I push changes 
This scenario produces the wrong version as the last recorded commit is actually the second last commit. It holds true even for this PR. Last commit hash is `fbe3f18` however last recorded version commit in `public/version.json` is `97c0788`.  We should think this through more carefully as this might provide a lot of false positives (eg. OSIM shows it runs on commit N-1 but actually it runs on commit N.

JakubFrejlach (changes requested):
I am not sure about the version update, this might create a lot of off by one errors. Some validation logic should be probably put in place.

src/stores/osimRuntime.ts
superbuggy: resolving after reimplementing

d10n:
Rebasing

scripts/run-local.sh
```diff
@@ -14,4 +14,7 @@ docker="$(command -v podman || command -v docker)"
 set -x
 
 "$docker" build -f local.Dockerfile -t osim:local
-"$docker" run --rm -p 8080:8080 --env OSIM_RUNTIME='{"backends":{"osidb":"/osidb","bugzilla":"http://bugzilla-service:8001"},"osimVersion":"0"}' --env OSIDB_BACKEND="$(cat ./public/runtime.json | jq -r .backends.osidb)" osim:local
+"$docker" run --rm -p 8080:8080 --env OSIM_ENV="dev" --env OSIM_BACKENDS_OSIDB="http://osidb-service:8000" --env OSIM_BACKENDS_BUGZILLA="http://bugzilla-service:8001" --env OSIM_BACKENDS_JIRA="http://jira-service:8002" --env OSIDB_BACKEND="$(cat ./public/runtime.json | jq -r .backends.osidb)" osim:local
```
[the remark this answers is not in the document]
superbuggy: Oh wait, I see the change to the general Dockerfile.

build/entrypoint.d/20-osim-runtime-json.sh
```diff
@@ -7,11 +7,33 @@
...
+  "env": "${OSIM_ENV}",
+  "backends": {
+    "osidb": "${OSIM_BACKENDS_OSIDB}",
+    "osidbAuth": "${OSIM_BACKENDS_OSIDB_AUTH}",
+    "bugzilla": "${OSIM_BACKENDS_BUGZILLA}",
+    "jira": "${OSIM_BACKENDS_JIRA}"
+  },
+  "osimVersion": ${OSIM_VERSION}
+}
+EOF
```
superbuggy, line 35: Nice

src/stores/osimRuntime.ts
```diff
@@ -23,29 +23,30 @@ export const osimRuntimeStatus = computed<OsimRuntimeStatus>(() => {
...
+  env: '',
+  backends: { osidb: '', osidbAuth: '', bugzilla: '', jira: '' },
+  osimVersion: { rev: '', tag: '', timestamp: '', dirty: true },
   error: '',
 });
-export const osimRuntime = computed<OsimRuntime>(() => {
-  return runtime.value;
-});
-
+export const osimRuntime = readonly(runtime.value);
```
JakubFrejlach, line 49: Also changing this to readonly creates two problems which I've encountered:

1. when changing osimRuntime to readonly, access to it changes from `osimRuntime.value.xxx` to just `osimRuntime.xxx`, there are several occurrences of that throughout the code
2. changing this to readonly produces osimRuntime with all values empty and the fetch from the runtime json file no longer mutates this  so whenever I access the osimRuntime in some different component I get empty strings I suppose this is not intended?

JakubFrejlach (changes requested):
Found some breaking problems with readonly.

src/stores/osimRuntime.ts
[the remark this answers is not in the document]
superbuggy: Ah looks like David merged in the suggestion, I've pushed a fix to supply the dependency.

superbuggy: It is not intended, I'd provided a code suggestion that contained non-functioning code that did not account for the API difference between `computed` and `readonly`. Thank you for your attentiveness, Jakub 👊

JakubFrejlach (approved):
LGTM now!

src/stores/osimRuntime.ts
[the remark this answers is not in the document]
d10n: I tried clicking the "commit suggestion" button in GitHub, not remembering that it needs to be imported, whoops.

PR #105, opened 2024-03-15
OSIDB-2306: Add validation function for embargoed and unembargoed dates

alizard0:
It implements [OSIDB-2306](https://issues.redhat.com/browse/OSIDB-2306)


src/components/FlawForm.vue
```diff
@@ -84,7 +84,8 @@ const errors = {
   cwe_id: null,
   major_incident_state: null,
   reported_dt: null,
-  unembargo_dt: null,
+  embargo_old_public_dt: "Set the current or a future date for an embargoed flaw.",
+  unembargo_future_public_dt: "Set the current or an old date for unembargoed flaw.",
```
JakubFrejlach, line 88: "Old date" sounds weird in this context, I would be for earlier/past/previous, wdyt?

src/components/FlawForm.vue
```diff
@@ -97,6 +98,22 @@ const flawCvss3CaculatorLink = computed(
...
+  if (flaw.embargoed) {
+    // if embargoed and updated date is older than now, it shows an error to set the current date or a future date instead
+    if (DateTime.fromISO(flaw.unembargo_dt as string).toISODate() < DateTime.now().toISODate()) {
+      return errors.embargo_old_public_dt;
+    }
+  } else {
+    // if NOT embargoed and updated date is in the future, it shows an error, to set the current date or an older date instead
+    if (DateTime.fromISO(flaw.unembargo_dt as string).toISODate() > DateTime.now().toISODate()) {
+      return errors.unembargo_future_public_dt;
+    }
```
JakubFrejlach, line 112: There is also one additional case which is apparently not mentioned in the Jira, but should be probably added. When Flaw is not embargoed and the unembargo_dt is not set, this is also invalid case and user should be prompted to set the unembargo_dt correctly. Also just a note, all these things are of course validated on the OSIDB side, so this servers merely as a convenience for user so he can observe these errors earlier.

JakubFrejlach (changes requested):
Some suggestions. Also these changes should be rebased on top of the current main as linter should run on PR.

JakubFrejlach:
Also I am not sure about the error message in this case:
image

Error message kinda blends in with the unembargo button, maybe we can adjust the colour a little bit? :thinking: Or add some visual borders?

src/components/FlawForm.vue
alizard0: Right, I agree.

src/components/FlawForm.vue
alizard0: What happens when is embargoed and dt is not set?

I have just pushed this commit:

https://github.com/RedHatProductSecurity/osim/pull/105/commits/4c26ecd4d451f2fd695dfc844f4a78281918831f#diff-bb5d5f91b338ff8b35ca76429e074f4af4afcd509d5adf9d6004016b419fd6cdR108

https://github.com/RedHatProductSecurity/osim/pull/105/commits/4c26ecd4d451f2fd695dfc844f4a78281918831f#diff-967a205363234f686e2f6df964ed2631bc6e26c2ba4f9d69abad87898d0d3461R443

Does it solves the use case?

src/components/FlawForm.vue
```diff
@@ -101,6 +102,27 @@ const onReset = () => {
   flaw.value = deepCopyFromRaw(initialFlaw.value as Record<string, any>) as ZodFlawType;
 };
 
+const validateFlawEmbargoDates = computed(
+  () => {
+    let unembargo_dt = DateTime.fromISO(String(flaw.value.unembargo_dt)).toISODate();
```
tboyetteredhat, line 107: Please use const instead of let.

tboyetteredhat (approved):

tboyetteredhat:
@alizard0 Please fix test failures. https://github.com/RedHatProductSecurity/osim/actions/runs/8390059868/job/22977542203?pr=105#step:6:3983
image

src/components/__tests__/FlawForm.spec.ts
```diff
@@ -9,10 +9,7 @@ import LabelEditable from '@/components/widgets/LabelEditable.vue';
 import IssueFieldStatus from '@/components/IssueFieldStatus.vue';
 import FlawForm from '../FlawForm.vue';
 import { useRouter } from 'vue-router';
-import LabelDiv from '../widgets/LabelDiv.vue';
-import LabelSelect from '../widgets/LabelSelect.vue';
```
tboyetteredhat, line 13: I think these lines are needed.

tboyetteredhat (changes requested):

src/components/__tests__/FlawForm.spec.ts
alizard0: It was a merge issue? Let me review it.

tboyetteredhat (approved):

src/components/FlawForm.vue
```diff
@@ -87,7 +87,8 @@ const errors = {
   cwe_id: null,
   major_incident_state: null,
   reported_dt: null,
-  unembargo_dt: null,
+  embargo_old_public_dt: 'Set the current or a future date for an embargoed flaw.',
+  unembargo_future_public_dt: 'Set the current or an previous date for unembargoed flaw.',
```
C-Valen, line 91: I think it would be simpler to return the strings from the validation method.

C-Valen (commented):
IMO we should avoid overlapping validation messages:
image

I think it could be better something like  [ExampleValidatedForm.vue](https://github.com/RedHatProductSecurity/osim/blob/main/src/components/ExampleValidatedForm.vue)
image

Additionally not sure if the validation can be made with zod for this logic, in which case I'd say that could be preferred.

src/components/FlawForm.vue
JakubFrejlach: No, I don't think so.
Flaw is embargoed and unembargo_dt is not set -> valid state
Flaw is not embargoed and unembargo_dt is not set -> invalid state

JakubFrejlach (commented):
Added some comments. Also I don't think it is necessary to include usernames in the commit names, it does not add much of a value and it clutters the git history.

src/components/FlawForm.vue
alizard0: I just defined them in the _const errors_ variable because I saw it was already defined.

alizard0:
@JakubFrejlach I covered the missing use case.
@C-Valen I changed LabelEditable warning to match the LabelInput as you suggested.

superbuggy:
Hey @alizard0,  #124 potentially covers the functionality here--the feature's scope evolved a bit to encompass this work. There we are using the `zod` library to perform the validations. We could huddle perhaps on Monday to look at how that works

C-Valen:
> Hey @alizard0, #124 potentially covers the functionality here--the feature's scope evolved a bit to encompass this work. There we are using the `zod` library to perform the validations. We could huddle perhaps on Monday to look at how that works

May be works better tuesday (or any other day) I think monday is holiday on some countries.
I am also interested on that huddle for zod validations.

alizard0:
Close PR because it overlaps with https://github.com/RedHatProductSecurity/osim/pull/124 and we agreed to merge the changes developed here into the other PR.

PR #108, opened 2024-03-18
Fixes advanced search and flaw editing

superbuggy:


C-Valen (approved):

JakubFrejlach (approved):
LGTM! Can we merge this? Broken flaw creation kinda blocks me and others probably as well, so it would be super useful.

src/types/zodFlaw.ts
```diff
@@ -149,7 +149,7 @@ export const ZodFlawCommentSchema = z.object({
     creation_time: z.string().nullish(),
     private_groups: z
       .string()
-      .transform((jsonString: string) => JSON.parse(jsonString))
+      .transform((jsonString: string) => JSON.parse(jsonString.replace(/'/g, '"')))
```
d10n, line 152: Wouldn't this create invalid JSON?

superbuggy:
```json
{
  "key": ["yes this would create 'invalid JSON'"]
}
```

superbuggy: In the case of the groups, they seem to be store in the following manner: `"['group_name','other_group_name']"`, but with the way I've implemented this fix, it will fail if that is not the case.

d10n: OK good, just checking for consideration. Deferring to @JakubFrejlach's approval

superbuggy: It may make sense to just remove this line, testing now...

JakubFrejlach: Actually on second thought, I checked the data and not every FlawComment does have the `private_groups` this might be the problem why is it parsed incorrectly?

PR #110, opened 2024-03-19
Update features for CWE ID update

qduanmu:


features/pages/flaw_detail_page.py
```diff
@@ -186,6 +189,17 @@ def set_input_field(self, field, value):
...
+    def get_input_value(self, field):
+        field_value = getattr(self, field + 'Value')
+        self.driver.execute_script("arguments[0].scrollIntoView(true);", field_value)
+        return field_value.get_text()
+
+    def clear_input_field(self, field):
+        field_btn = field + 'EditBtn'
+        self.click_btn(field_btn)
+        field_input = getattr(self, field + 'Input')
+        field_input.send_keys(Keys.CONTROL + 'a', Keys.BACKSPACE)
```
RichardXuan, line 201: Why not use `execute_script("arguments[0].value = '';")`, not working?

RichardXuan (approved):
LGTM

features/pages/flaw_detail_page.py
huiwangredhat: I also met the same issue. 
After using execute_script, add time.sleep(2), it works.

PR #111, opened 2024-03-19
Lint reenable

JakubFrejlach:
This PR reenables linting on build so we can run linter via github workflows. I am not sure why the linter got disabled in the first place, probably some changes overwritten that.


tboyetteredhat (approved):

package.json
```diff
@@ -20,7 +20,7 @@
     "vibe-check": "npx vite-bundle-visualizer",
     "generate-openapi-client": "openapi-generator-cli generate -i osidb-3.4.0.oas3.json --output src/generated-client -g typescript-axios",
     "mock-osidb-server": "prism mock osidb-3.0.0.oas3.json",
-    "lint": "eslint . --ext .js,.ts,.vue"
+    "lint": "eslint . --ext .js,.ts,.vue --max-warnings=0"
```
superbuggy, line 23: In my mind, the warnings give valuable feedback and my preference would be to keep them

superbuggy (changes requested):

package.json
JakubFrejlach: Everything is going to be preserved, this is just a way how to treat every warning as error in the eslint, it basically just tells the linter that any warning is unacceptable and linter returns the error code.

JakubFrejlach: Ok, I understand, so probably lets remove this and declare some things as errors as I've suggested? I was thinking at least about indentation as the inconsistencies on that are sometimes really crazy, wdyt?

superbuggy: That sounds great

d10n (approved):

JakubFrejlach:
@superbuggy addressed your suggestions

superbuggy (approved):

PR #112, opened 2024-03-19
Defines read only mode behavior

superbuggy:


src/stores/osimRuntime.ts
```diff
@@ -27,21 +27,29 @@ const OsimRuntime = z.object({
...
-  osimVersion: z.string(),
+  osimVersion: z.object({
+    rev: z.string(),
+    tag: z.string(),
+    timestamp: z.string(),
+    dirty: z.boolean(),
+  }),
+  env: z.string().default(''),
   error: z.string().default(''),
+  readOnly: z.boolean().default(false),
```
JakubFrejlach, line 38: [docs](https://github.com/RedHatProductSecurity/osim/blob/feature/OSIDB-2317-read-only-mode/docs/overview.md#local-configuration) local config example should be refined to comply with the changes.

src/stores/osimRuntime.ts
```diff
@@ -92,22 +112,26 @@ function fetchRuntime() {
         runtime.value.error = 'Backends are not correctly configured. Please try again later.';
         status.value = OsimRuntimeStatus.ERROR;
       }
+      if (runtime.value.osimVersion.timestamp === '1970-01-01T00:00:00Z') {
+        // 0 unix timestamp
+        runtime.value.osimVersion.timestamp = new Date().toISOString();
+      }
```
JakubFrejlach, line 118: What is the purpose of this condition?

JakubFrejlach (changes requested):
Some observations and suggestions.

JakubFrejlach:
@superbuggy Also I've noticed that this branch is 10 commits behind the `main` but it does not reports any conflicts, which is really weird, because at least the `package.json` file removes again the lint functionality from the build command, there should be conflict reported at least. :thinking:

superbuggy:
> @superbuggy Also I've noticed that this branch is 10 commits behind the `main` but it does not reports any conflicts, which is really weird, because at least the `package.json` file removes again the lint functionality from the build command, there should be conflict reported at least. 🤔

It is common for feature branches to lag behind main, and in this case since that conflict was already resolved and there are no commits _unique to this branch_ to `package.json`, there are no conflicts, so it is expected that no conflicts be reported.

It wasn't clear if this was the change you were requested, if so, then no change is necessary as there is no conflict.

src/stores/osimRuntime.ts
JakubFrejlach: I see. Thank you.

JakubFrejlach:
> > @superbuggy Also I've noticed that this branch is 10 commits behind the `main` but it does not reports any conflicts, which is really weird, because at least the `package.json` file removes again the lint functionality from the build command, there should be conflict reported at least. 🤔
> 
> It is common for feature branches to lag behind main, and in this case since that conflict was already resolved and there are no commits _unique to this branch_ to `package.json`, there are no conflicts, so it is expected that no conflicts be reported.
> 
> It wasn't clear if this was the change you were requested, if so, then no change is necessary as there is no conflict.

My concern was that the linting is not yet again erased from the package.json scripts however you've already rebased that. My change request was more related to documenting the new example of runtime.json.

JakubFrejlach (approved):

C-Valen (approved):
Huddle reviewed. LGTM.

tboyetteredhat (approved):

d10n (approved):

PR #115, opened 2024-03-20
Fixes refresh, fixes post-create redirect, adds cancellation to new references/acknowledgments

superbuggy:


src/composables/useFlawModel.ts
```diff
@@ -44,18 +39,21 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), emit: FlawEmitt
   const osimLink = computed(() => getFlawOsimLink(flaw.value.uuid));
 
   async function createFlaw() {
-    postFlaw(flaw.value)
+    isSaving.value = true;
```
C-Valen, line 42: Isn't `isSaving` being handled twice for update/create actions, as it is still on `FlawForm` `onSubmit`?

src/composables/useFlawModel.ts
```diff
@@ -44,18 +39,21 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), emit: FlawEmitt
...
       .then((response: any) => {
         router.push({
-          name: 'flaw-detail',
-          params: { id: response?.uuid },
+          name: 'flaw-details',
+          params: { id: response?.cve_id || response?.uuid },
         });
       })
       .catch(createCatchHandler('Error creating Flaw'));
+    isSaving.value = false;
```
C-Valen, line 52: Not sure about this, but I'm thinking on the proper way to ensure `isSaving = false` is set in every scenario.
```suggestion
      .catch(createCatchHandler('Error creating Flaw'))
      .finally(() => { isSaving.value = false; });
```

src/composables/useFlawModel.ts
```diff
@@ -44,18 +39,21 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), emit: FlawEmitt
...
   }
 
   async function updateFlaw() {
+    isSaving.value = true;
```
C-Valen, line 56: Same case a above:
> Isn't isSaving being handled twice for update/create actions, as it is still on FlawForm onSubmit?

C-Valen (commented):
Generally looks good, just small details that may be worth considering.

src/components/IssueFieldAcknowledgments.vue
```diff
@@ -45,10 +46,18 @@ function handleDelete(uuid: string, closeModal: () => void) {
...
-            from
-            <input v-model="items[itemIndex].affiliation" type="text" />
+          <div class="d-flex justify-content-between">
+            <div>
+              <input v-model="items[itemIndex].name" type="text" />
+              from
+              <input v-model="items[itemIndex].affiliation" type="text" />
+            </div>
+            <div
+              class="text-end  osim-cancel-new-acknowledgment"
```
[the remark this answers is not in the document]
superbuggy: No :( I have a separate PR for this #125

src/composables/useFlawModel.ts
superbuggy: Yes, thank you!

src/composables/useFlawModel.ts
superbuggy: Yes, I can ditch the await above.

C-Valen (approved):
🚀

PR #116, opened 2024-03-21
Task/OSIDB-2313 FlawForm Modifications

C-Valen:
Implements **OSIDB-2313** and applies changes based on stakeholder feedback:
- New layout on Descriptions, References and acknowledgements
- Renames `Description` (Comment#0) and `Summary` based on OSIDB API logic
- Makes "description" text area fields visible with empty content (hidden/shown after saving flaw)
- Adds optional behavior control on `LabelCollapsable` when passing `isExpandable` property
- Modifies `EditableList` to use new `isExpandable` property of `LabelCollapsable`
    - Disables `EditableList` collapse button if no items to list
- Shows number of items on `EditableList` label
- Rises the spacing between `FlawForm` sections
- Fixes overlapping on edit buttons of `EditableList` items
- Other minor style adjustments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 ## [Unreleased]
 ### Changed
 * Flaw URLs consists preferably of CVE ID if possible (OSIDB-2018)
+* Changed layout for Descriptions, References and Acknowledgements
+* Renamed `Description` (to Comment#0) and `Summary` (to Description)
+* Made text area fields visible with empty content
+* Minor style adjustments on `Flaw` views
+
+### Fixed
+* Fixed overlap on edit buttons at References and Acknowledgements items
+* Fixed inconsistent style on read-only fields with empty value
 
 ## [2024.1.0]
 
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -246,59 +246,62 @@ const cvssString = computed(() => {
           <LabelEditable v-model="flaw.team_id" type="text" label="Team ID" />
         </div>
       </div>
+      <div class="mt-3 pt-4 pb-3 mb-4 border-top">
+        <LabelTextarea v-model="flaw.description" label="Comment#0" />
+        <LabelTextarea v-if="hasSummary" v-model="flaw.summary" label="Description" />
+        <LabelTextarea v-if="hasStatement" v-model="flaw.statement" label="Statement" />
+        <LabelTextarea v-if="hasMitigation" v-model="flaw.mitigation" label="Mitigation" />
+        <div class="d-flex gap-3 mb-3">
+          <button
+            v-if="!flaw.description"
+            class="btn btn-secondary"
+            @click="flaw.description = 'Description Text ...'"
+          >
+            Add Description
+          </button>
+          <button
+            v-if="!flaw.statement"
+            class="btn btn-secondary"
+            @click="flaw.statement = 'Statement Text ...'"
+          >
+            Add Statement
+          </button>
+          <button
+            v-if="!flaw.mitigation"
+            class="btn btn-secondary"
+            @click="flaw.mitigation = 'Mitigation Text ...'"
+          >
+            Add Mitigation
+          </button>
+        </div>
+      </div>
       <div class="mt-3 pt-4 pb-3 mb-4 border-top border-bottom">
-        <div class="osim-doc-text-container">
-          <LabelCollapsable label="Document Text Fields">
-            <LabelTextarea v-if="flaw.summary" v-model="flaw.summary" label="Summary" />
-            <LabelTextarea v-if="flaw.description" v-model="flaw.description" label="Description" />
-            <LabelTextarea v-if="flaw.statement" v-model="flaw.statement" label="Statement" />
-            <LabelTextarea v-if="flaw.mitigation" v-model="flaw.mitigation" label="Mitigation" />
-            <div v-if="!flaw.summary" class="mb-3">
-              <button class="btn btn-secondary" @click="flaw.summary = 'Summary Text ...'">
-                Add Summary
-              </button>
-            </div>
-            <div v-if="!flaw.description" class="mb-3">
-              <button class="btn btn-secondary" @click="flaw.description = 'Description Text ...'">
-                Add Description
-              </button>
-            </div>
-            <div v-if="!flaw.statement" class="mb-3">
-              <button class="btn btn-secondary" @click="flaw.statement = 'Statement Text ...'">
-                Add Statement
-              </button>
-            </div>
-            <div v-if="!flaw.mitigation" class="mb-3">
-              <button class="btn btn-secondary" @click="flaw.mitigation = 'Mitigation Text ...'">
-                Add Mitigation
-              </button>
-            </div>
-          </LabelCollapsable>
+        <div class="d-flex gap-3">
           <IssueFieldReferences
             v-model="flawReferences"
             :isEmbargoed="flaw.embargoed"
+            class="w-100"
             @reference:update="saveReferences"
             @reference:new="addBlankReference(flaw.embargoed)"
-            @reference:cancel-new="cancelAddReference"
             @reference:delete="deleteReference"
           />
           <IssueFieldAcknowledgments
             v-model="flawAcknowledgments"
+            class="w-100"
             @acknowledgment:update="saveAcknowledgments"
             @acknowledgment:new="addBlankAcknowledgment(flaw.embargoed)"
-            @acknowledgment:cancel-new="cancelAddAcknowledgment"
             @acknowledgment:delete="deleteAcknowledgment"
           />
-          <LabelCollapsable label="Trackers">
-            <ul>
-              <li v-for="(tracker, trackerIndex) in trackerUuids" :key="trackerIndex">
-                <RouterLink :to="{ name: 'tracker-details', params: { id: tracker.uuid } }">
-                  {{ tracker.display }}
-                </RouterLink>
-              </li>
-            </ul>
-          </LabelCollapsable>
         </div>
+        <LabelCollapsable label="Trackers">
+          <ul>
+            <li v-for="(tracker, trackerIndex) in trackerUuids" :key="trackerIndex">
+              <RouterLink :to="{ name: 'tracker-details', params: { id: tracker.uuid } }">
+                {{ tracker.display }}
+              </RouterLink>
+            </li>
+          </ul>
+        </LabelCollapsable>
       </div>
       <AffectedOfferings
         :theAffects="theAffects"
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -98,6 +98,10 @@ const errors = {
   source: null,
 };
 
+const hasSummary = ref(!!flaw.value.summary && flaw.value.summary.trim() != '');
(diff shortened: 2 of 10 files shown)


src/components/FlawForm.vue
```diff
@@ -241,69 +245,83 @@ const cvssString = computed(() => {
...
+        <LabelTextarea v-if="hasSummary" v-model="flaw.summary" label="Description" />
+        <LabelTextarea v-if="hasStatement" v-model="flaw.statement" label="Statement" />
+        <LabelTextarea v-if="hasMitigation" v-model="flaw.mitigation" label="Mitigation" />
+        <div class="d-flex gap-3 mb-3">
+          <button
+            v-if="!hasSummary"
+            class="btn btn-secondary"
+            @click="
+              hasSummary = true;
+              flaw.summary = 'Description Text ...';
```
d10n, line 259: Does placeholder="Description Text..." work on the textarea instead?

src/components/FlawForm.vue
```diff
@@ -93,6 +93,10 @@ const errors = {
   source: null,
 };
 
+const hasSummary = ref(!!flaw.value.summary && flaw.value.summary.trim() != '');
+const hasStatement = ref(!!flaw.value.statement && flaw.value.statement.trim() != '');
+const hasMitigation = ref(!!flaw.value.mitigation && flaw.value.mitigation.trim() != '');
```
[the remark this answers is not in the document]
superbuggy: Could these be computed instead of refs?

src/components/widgets/LabelStatic.vue
```diff
@@ -14,7 +14,7 @@ defineProps<{
         {{ label }}
       </span>
       <div :class="{ 'alert alert-warning': !modelValue }">
-        <span class="form-control">{{ modelValue }}</span>
+        <span class="form-control" style="border-top-left-radius: 0">{{ modelValue }}</span>
```
superbuggy, line 17: I might prefer to avoid using an inline style here

src/components/widgets/EditableList.vue
```diff
@@ -55,103 +56,107 @@ function commitEdit(index: number) {
...
-              v-if="indexBeingEdited === itemIndex"
-              type="button"
-              class="btn"
-              @click="cancelEdit(itemIndex)"
-            >
-              <i class="bi bi-x">
-                <span class="visually-hidden">Cancel {{ entityName }} Edit </span>
-              </i>
-            </button>
+  <div class="my-3">
```
superbuggy, line 59: I think we could just pass this class to LabelCollapsable and modify it to apply any classes passed in as attributes

src/components/FlawForm.vue
```diff
@@ -241,69 +245,89 @@ const cvssString = computed(() => {
...
-            <ul>
-              <li v-for="(tracker, trackerIndex) in trackerUuids" :key="trackerIndex">
-                <RouterLink :to="{ name: 'tracker-details', params: { id: tracker.uuid } }">
-                  {{ tracker.display }}
-                </RouterLink>
-              </li>
-            </ul>
-          </LabelCollapsable>
         </div>
+        <LabelCollapsable label="Trackers" :length="trackerUuids.length">
```
superbuggy, line 310:
```suggestion
        <LabelCollapsable :label="trackerUuids.length ? `Trackers: ${trackerUuids.length}` : 'Trackers'} `">
```

superbuggy: This approach seems simpler to me than adding a length proper to LabelCollapsable, but adding fallthrough for disabled makes sense, or adding an isExpanded prop.

src/components/widgets/LabelCollapsable.vue
```diff
@@ -28,12 +30,24 @@ function handleClick() {
...
+    >
       <i v-if="isExpanded" class="bi bi-dash-square-dotted me-1"></i>
       <i v-else class="bi bi-plus-square-dotted me-1"></i>
       <slot name="label">
         <label class="ms-2 form-label">
-          {{ label }}
+          {{ length > 0 ? `${label}: ${length}` : `${label}` }}
         </label>
       </slot>
     </button>
```
superbuggy, line 53: My original intent with this component was to be a replacement for details/summary, I think it might not need to know anything about its contents. 

However, I think it would make sense to either:
 1. Add a prop `isExpandable` to disable the expand button (gray it out, or hide it)
 2.  Use a `disabled` fall-through attribute to gray the button. This would be my preference because it would let the user know that usually a clickable button is there, just that it can't be clicked when there are no available trackers.

src/components/IssueFieldAcknowledgments.vue
```diff
@@ -22,33 +22,34 @@ function handleDelete(uuid: string, closeModal: () => void) {
...
         </div>
       </template>
 
       <template #edit-form="{ items, itemIndex }">
-        <div class="form-group">
-          <div>
-            <input v-model="items[itemIndex].name" type="text" />
-            from
-            <input v-model="items[itemIndex].affiliation" type="text" />
+        <div class="form-group mb-2" style="max-width: 85%">
```
superbuggy, line 38: I have a general preference to avoid inline styles

src/components/IssueFieldAcknowledgments.vue
```diff
@@ -22,33 +22,34 @@ function handleDelete(uuid: string, closeModal: () => void) {
...
         </div>
       </template>
 
       <template #create-form="{ items, itemIndex }">
-        <div class="form-group">
-          <div>
-            <input v-model="items[itemIndex].name" type="text" />
-            from
-            <input v-model="items[itemIndex].affiliation" type="text" />
+        <div class="form-group mb-2" style="max-width: 85%">
```
superbuggy, line 48: Same as prior comment

src/components/widgets/LabelStatic.vue
C-Valen: Buy it, is true that there is no inline styles in the code base so no reason to start using it.

src/components/widgets/EditableList.vue
C-Valen: The thing is that the form on `EditableList` (buttons to add/save new list items) was intended to wrapped by the vertical margin, adding the class to `LabelCollapsable` would exclude the form. 

But I see your point on not using strict styles on reusable components, so I think will be best to add that class on the specific components on `FlawForm` (`IssueFieldAcknowledgments` & `IssueFieldReferences`)

src/components/widgets/LabelCollapsable.vue
C-Valen: Totally agree, the length approach doesn't look clean to me now as well. 

I opted for adding `isExpandable` property on `LabelCollapsable` so false statements passed from parent components will disable the button.

IMO the fall-through limits the component's re-usability, as currently is being used only within lists, but we might want to use it in the future for non-list Labeled Collapsible.

src/components/widgets/EditableList.vue
superbuggy: Ah, I just meant to suggest adding the functionality to the component to apply any classes passed to the component, but not to modify the component itself, so we could optionally apply any classes that we would want. I think I agree re the components in FlawForm, if I follow correctly.

src/components/widgets/LabelTextarea.vue
```diff
@@ -40,6 +40,7 @@ const elTextArea = ref();
         :ref="elTextArea"
         class="form-control col-9 d-inline-block"
         :class="{ 'is-invalid': error != null }"
+        :placeholder="placeholder"
```
d10n, line 43: I think this is handled by line 39 `v-bind="$attrs"` but no harm in adding this explicitly

src/components/widgets/LabelStatic.vue
```diff
@@ -13,16 +13,21 @@ defineProps<{
...
       </div>
     </div>
     <div v-else class="row">
       <span class="form-label col-3">
         {{ label }}
       </span>
-      <div class="col-9" :class="{ 'alert alert-warning': !modelValue }">
-        <span class="form-control">{{ modelValue }}</span>
+      <div class="col-9">
+        <span class="form-control" :class="{ 'alert alert-warning': !modelValue }">{{ modelValue }}</span>
```
JakubFrejlach, line 30: Good job on addressing this one! It was bugging my mind in my branch and was planning to address it afterwards :smile:

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
...
 ### Changed
 * Flaw URLs consists preferably of CVE ID if possible (OSIDB-2018)
+* Changed layout on Descriptions, References and Acknowledgements
+* Renamed `Description` (Comment#0) and `Summary` (Description) based on OSIDB API logic
+* Made text area fields visible with empty content
+* Minor `FlawForm` style adjustments
+
+### Fixed
+* Fixed overlapping on edit buttons of `EditableList` items
+* Fixed inconsistent style on `LabelStatic` when model value is empty
```
JakubFrejlach, line 13: I think changelog's main focus should be user facing changes and thus they should be probably described with "non-developer" vocabulary. Eg. terms like `LabelStatic` or `EditableList` are not very suitable as user might (and probably will) have no idea what is that referring to.

JakubFrejlach (approved):
One small remark on changelog message handling, otherwise LGTM!

tboyetteredhat (approved):
LGTM. Nice job 😄

src/components/FlawForm.vue
```diff
@@ -98,6 +98,10 @@ const errors = {
   source: null,
 };
 
+const hasSummary = ref(flaw.value.summary && flaw.value.summary.trim() !== '');
+const hasStatement = ref(flaw.value.statement && flaw.value.statement.trim() !== '');
+const hasMitigation = ref(flaw.value.mitigation && flaw.value.mitigation.trim() !== '');
+
```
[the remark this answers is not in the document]
C-Valen: That was my initial approach, but using `computed()` is not working as I expected here. It makes the field buttons to submit the whole `FlawForm` when actioned, instead than just make the fields visible (don't know why).
I tried to control the event on the buttons but I couldn't find a workaround, after trying using refs and seeing that it works as expected I decided to keep it like this.
Do you think can be any side harm for using refs here @superbuggy ?

CHANGELOG.md
C-Valen: That has lot of sense, I agree. Thanks for clarifying.

src/components/widgets/LabelTextarea.vue
C-Valen: Amended on [🗑️ Remove redundant placeholder bind](https://github.com/RedHatProductSecurity/osim/pull/116/commits/f2da86ba15b7ca3b20273f419e8ab0dd9ad08ae5)

src/components/FlawForm.vue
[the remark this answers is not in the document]
C-Valen: For the record, just adding `type="button"` on the buttons prevented to trigger the parent event (form submit for this case). Thanks @superbuggy for the suggestion off thread
However the properties hasSummary, hasStatement & hasMitigation are being declared as refs cause the values need to be updated on runtime. At least from the approach's perspective I have on this they should be initialized based in the existence of content for each associated model value, but after they are changing value from the corresponding button events. 
Not sure if there could be more optimal way to do this, but using computed will require defining getter/setters for each of them which I think is less popper.

superbuggy (approved):

PR #120, opened 2024-03-22
OSIDB-1207: compare CVSS differences

tboyetteredhat:


src/components/FlawForm.vue
```diff
@@ -102,14 +103,41 @@ const onReset = () => {
...
+    const charFromCvss = i < cvssValue.length ? cvssValue[i] : '';
+    if (charFromFlaw === charFromCvss) {
+      result += charFromFlaw;
+    } else {
+      result += highlightScore(charFromFlaw);
+    }
+  }
+
+  return result;
 });
```
superbuggy, line 135:
```suggestion
const highlightedNvdCvssScore = computed(() => {
  let result = [];
  const flawValue = `${flawNvdCvssScore.value}`;
  const cvssValue = `${cvssString.value}`;

  const maxLength = Math.max(flawValue.length, cvssValue.length);

  for (let i = 0; i < maxLength; i++) {
    const charFromFlaw = i < flawValue.length ? flawValue[i] : '';
    const charFromCvss = i < cvssValue.length ? cvssValue[i] : '';
    result.push({ char: charFromFlaw, isHighlighted: charFromFlaw !== charFromCvss });
  }

  return result;
});
```

src/components/FlawForm.vue
```diff
@@ -177,9 +205,13 @@ const cvssString = computed(() => {
           <LabelInput v-model="flawRhCvss.score" label="CVSSv3 Score" type="text" />
           <div class="row">
             <div :class="['col', { 'cvss-button-div': displayCvssNISTForm }]">
-              <LabelStatic v-model="flawNvdCvssScore" label="NVD CVSSv3" type="text" />
+              <LabelStaticHighlighted 
+                v-model="highlightedNvdCvssScore" 
+                label="NVD CVSSv3" 
+                type="text" 
+              />
```
superbuggy, line 212:
```suggestion
              <LabelDiv label="NVD CVSSv3">
                <span
                  v-for="char in highlightedNvdCvssScore"
                  :key="char.char"
                  :class="{'text-primary': char.isHighlighted}"
                >
                  {{ char }}
                </span>
              </LabelDiv>
```

superbuggy (commented):
This is awesome Tyler. I have one suggestion to avoid using sanitizeHtml and adding another component, but totally up to you.

src/components/FlawForm.vue
```diff
@@ -102,14 +103,41 @@ const onReset = () => {
...
 const cvssString = computed(() => {
-  return `${flawRhCvss.value?.score}/${flawRhCvss.value?.vector}`;
+  const values = [flawRhCvss.value?.score, flawRhCvss.value?.vector].filter(value => !!value);
+  return values.join('/');
+});
+
+const highlightedNvdCvssScore = computed(() => {
+  let result = '';
+  const flawValue = String(flawNvdCvssScore.value);
+  const cvssValue = String(cvssString.value);
```
superbuggy, line 117: If these values will always be strings, so the casting isn't necessary, unless they are null or undefined sometimes. We could adjust this in the zod schema to give them default values of `''` with `.default('')` and then we would have the benefit of knowing they would always be strings... Thoughts?

tboyetteredhat: I think the problem is that  `nistCvssScore` can be undefined and flawNvdCvssScore can be undefined. so i think we need to do like this.

`
  const nvdCvssString = computed(() => {
    const values = [nistCvssScore.value?.score, nistCvssScore.value?.vector].filter(value => !!value);
    return values.join('/') || flaw.value.nvd_cvss3 || '-';
  });
`

superbuggy (approved):
Looks great, Tyler!

C-Valen (approved):

PR #124, opened 2024-03-26
Validates Flaw Fields with Visual Feedback 

superbuggy:


C-Valen (commented):
Code looks good. I really like the toast with list of validation issues.

I'd suggest updating `EditableDate` and `EditableText` to avoid error messages overlaping on the form fields. (Also mentioned on https://github.com/RedHatProductSecurity/osim/pull/105#pullrequestreview-1962992344)

However for this case I think that showing same validation messages twice can be kind of redundant. My preference would be to keep the toast for specific details of the validation and in the form just mark the related fields (like `Public Date` is currently doing).
image

Would be cool to use that style in red for required fields not filled, and in yellow for fields that failed in the validation. Or something like that.

Just my thoughts, totally debatable.

superbuggy:
> I'd suggest updating `EditableDate` and `EditableText` to avoid error messages overlaping on the form fields. (Also mentioned on [#105 (review)](https://github.com/RedHatProductSecurity/osim/pull/105#pullrequestreview-1962992344))

I noticed the same thing and wanted to huddle about the UI convention, about how to handle the error message indicators in the UI. Should we just show a hoverable icon that shows the error message in the UI?
 
> However for this case I think that showing same validation messages twice can be kind of redundant. My preference would be to keep the toast for specific details of the validation and in the form just mark the related fields (like `Public Date` is currently doing). 

I agree--it could be nice to have the feedback as the user types. What do you think about the hoverable error icon idea above?

C-Valen:
> I agree--it could be nice to have the feedback as the user types. What do you think about the hoverable error icon idea above?

The hoverable icon is a very good idea, it could be a clean way to display the info/messages.
I definitely agree on a huddle to define some UI conventions.

src/components/FlawComments.vue
```diff
@@ -30,12 +30,18 @@ function parseGroups(serializedJson: string) {
   }
 }
 
+function parseIsPrivate(isPrivate: string) {
```
alizard0, line 33: Wouldnt isParsePrivate be a better name?

superbuggy: The purpose is this function is to parse the field `is_private`

superbuggy: If we had an individual component (`FlawComment` singular) for each comment and passed in that `comment.meta-attr` data object as a prop, I would probably use a computed property instead of this function and call it `isPrivate`

alizard0 (approved):
LGTM

tboyetteredhat (approved):

C-Valen (approved):
lgtm

PR #125, opened 2024-03-26
Provides formatting rules

superbuggy:


.gitignore-blame-revs
d10n, line 1: Rename the file to `.git-blame-ignore-revs` ([GH default filename](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view), also mentioned on line 2)

C-Valen (approved):
Nice one! Good work!

PR #130, opened 2024-03-27
OSIDB 2320: Remove widget test page

C-Valen:
Applies changes of **OSIDB-2320**:
- Removes widget test page option from `Navbar` component (except for dev env)
- Controls route access to `/widget-test` allowing access only in dev env
- Adds a test case on `Navbar.spec.ts` to ensure widget test is hidden


tboyetteredhat (approved):
LGTM

PR #133, opened 2024-03-28
Fix MyIssues filter on IndexView

tboyetteredhat:
MyIssues filter fix on IndexView

After user login, MyIssues filter is passing 'Current_user' filter to owner query. This is happening because of losing reactivity on username.


superbuggy (approved):

superbuggy:
Great catch, Tyler

tboyetteredhat:
> Great catch, Tyler

Thanks for resolving merge conflict.

PR #135, opened 2024-03-28
OSIDB-2308: Readonly Comment#0 and CVSS3 score

JakubFrejlach:
This PR:
* converts CVSS3 score to readonly fields
* makes Flaw Comment#0 writable only upon Flaw creation

Partially addresses OSIDB-2308

### DIFF
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -263,10 +263,11 @@ const onReset = () => {
           label="Comment#0"
           placeholder="Comment#0 ..."
           :error="errors.description"
+          :disabled="mode === 'edit'"
         />
         <LabelTextarea
           v-if="showSummary"
-          v-model="flaw.summary" 
+          v-model="flaw.summary"
           label="Description"
           placeholder="Description Text ..."
           :error="errors.summary"
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -7,7 +7,7 @@ import { deepCopyFromRaw } from '@/utils/helpers';
 import LabelEditable from '@/components/widgets/LabelEditable.vue';
 import LabelSelect from '@/components/widgets/LabelSelect.vue';
 import LabelTextarea from '@/components/widgets/LabelTextarea.vue';
-import LabelInput from '@/components/widgets/LabelInput.vue';
+import LabelStatic from '@/components/widgets/LabelStatic.vue';
 import LabelCollapsable from '@/components/widgets/LabelCollapsable.vue';
 import AffectedOfferings from '@/components/AffectedOfferings.vue';
 import IssueFieldEmbargo from '@/components/IssueFieldEmbargo.vue';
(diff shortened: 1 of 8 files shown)


JakubFrejlach:
Basic functionality implemented, tests need to be fixed.

C-Valen (commented):
An observation on the Comment#0 read-only approach, it might not look ideal to me from an UX perspective, as it still feels like an editable field cause of the highlights when focused. I think it could be confusing for some users.
I'm not sure if would be possible having like a a "LabelStaticTextArea" (for like having equivalent implementations in both fields) or just making it non focusable when readonly is true.
Just a suggestion, the main functionality is totally covered and solid.

src/components/FlawForm.vue
```diff
@@ -264,7 +263,12 @@ const onReset = () => {
         </div>
       </div>
       <div class="osim-flaw-form-section border-top">
-        <LabelTextarea v-model="flaw.description" label="Comment#0" placeholder="Comment#0 ..." />
+        <LabelTextarea
+          v-model="flaw.description"
+          label="Comment#0"
+          placeholder="Comment#0 ..."
+          :readonly="mode === 'edit' ? true : false"
```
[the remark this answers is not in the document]
JakubFrejlach: Agreed!

C-Valen:
> > An observation on the Comment#0 read-only approach, it might not look ideal to me from an UX perspective, as it still feels like an editable field cause of the highlights when focused. I think it could be confusing for some users. I'm not sure if would be possible having a `LabelStaticTextArea` (for equivalent implementations in both fields) or just making it non focusable when readonly is true. Just a suggestion, the main functionality is totally covered and solid.
> 
> There is also a possibility to make the field disabled, it shows the text but it doesn't appear to be editable. 🤔

Sure! that's also a good and simpler idea.

JakubFrejlach:
> > > An observation on the Comment#0 read-only approach, it might not look ideal to me from an UX perspective, as it still feels like an editable field cause of the highlights when focused. I think it could be confusing for some users. I'm not sure if would be possible having a `LabelStaticTextArea` (for equivalent implementations in both fields) or just making it non focusable when readonly is true. Just a suggestion, the main functionality is totally covered and solid.
> > 
> > 
> > There is also a possibility to make the field disabled, it shows the text but it doesn't appear to be editable. 🤔
> 
> Sure! that's also a good and simpler idea.

Implemented. Can you re-review please?

src/components/widgets/LabelTextarea.vue
```diff
@@ -4,11 +4,16 @@ import { ref } from 'vue';
...
   modelValue: string | null | undefined;
   label: string;
   error?: string;
-}>();
+  disabled?: boolean;
+}>(), {
+  error: '',
+  readonly: false
+});
+
```
C-Valen, line 16: Why not adding `disabled` prop with defaults?

C-Valen (approved):
Looks good! 
I really like the style on disabled state on `Comment#0` it distinct clearly from rest of editable fields. Good job!

src/components/widgets/LabelTextarea.vue
JakubFrejlach: That's because I meant to rename the property in defaults from `readonly` to `disabled` as well and didn't do it :smile: Thanks!

tboyetteredhat (approved):
LGTM.
I like styling on comment#0

superbuggy:
What if Comment#0 has a typo, it couldn't be amended after creation? If that's desirable then 👍

PR #138, opened 2024-04-02
add test case: Add RHSB reference with incorrect link

RichardXuan:


features/constants.py
```diff
@@ -6,3 +6,4 @@
 BUGZILLA_API_KEY = os.getenv("BUGZILLA_API_KEY")
 JIRA_API_KEY = os.getenv("JIRA_API_KEY")
 TIMEOUT = '10'
+TEST_FLAW_CVE_ID = os.getenv("TEST_FLAW_CVE_ID")
```
huiwangredhat, line 9: I guess we need several different CVE_IDs for different tests, so how about setting the cve_id as a param?

huiwangredhat (approved):
LGTM.

PR #141, opened 2024-04-03
✨ Provide changelog modal

superbuggy:


public/CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
+# OSIM Changelog
```
JakubFrejlach, line 1: Does this mean we now need to keep updating both `CHANGELOG.md` and `public/CHANGELOG.md`? If so, is there way how to either keep just this one or somehow provide a "link" to the one in a root? I mean, having two CHANGELOG files calls for inconsistencies.

src/components/ChangeLog.vue
```diff
@@ -0,0 +1,66 @@
...
+            @click="closeModal"
+          ></button>
+        </div> 
+  
+      </template>
+      <template #body>
+        <div v-html="html.body" />
+      </template>
+      <template #footer>
+        <button class="btn btn-primary" @click="closeModal">Close</button>
```
JakubFrejlach, line 54: Isn't it redundant to have two differently looking closing buttons on one modal?

JakubFrejlach (approved):
Added suggestions and questions.

src/components/ChangeLog.vue
superbuggy: It is, but one of them you have to scroll to see and is more obvious if you miss the light grey x in the upper left corner

public/CHANGELOG.md
superbuggy: This is only in public for local development purposes--I change the content of the file to reflect that

superbuggy: Actually a symlink might be better here...

tboyetteredhat:
@superbuggy I think there's some ui issues. 
image

- Close icon is not showing right.
- Scroll behavior on modal
- - Modal header
- - Close button

C-Valen:
> @superbuggy I think there's some ui issues. image
> 
>     * Close icon is not showing right.
> 
>     * Scroll behavior on modal
> 
>     * * Modal header
> 
>     * * Close button

I thought the same as @tboyetteredhat.
Just pointing couple of suggestions:
- Move close icon top-right with `sticky`/`absolute` positioning as only closing option (being always visible).
- Add a bit of padding on body content

Overall the approach looks nice to me, considering the difficulties on having access to the file.

C-Valen (approved):
Nice one! Looks good 🔥 

One small consideration:
Isn't it more logic having the link on the footer, than in settings?

superbuggy:
> Nice one! Looks good 🔥
> 
> One small consideration: Isn't it more logic having the link on the footer, than in settings?

That had been my thought but then someone pointed out (@d10n, maybe) it would be too easy to miss and/or inconvenient? We could include it in both places I think

d10n (approved):

PR #143, opened 2024-04-04
OSIDB-2321: Removed trackers from create flaw form and added the unit tests too

alizard0:
It removes the trackers section from the FlawForm when it is on CREATE mode only.

Implements https://issues.redhat.com/browse/OSIDB-2321

### DIFF
--- a/src/components/__tests__/FlawForm.spec.ts
+++ b/src/components/__tests__/FlawForm.spec.ts
@@ -12,7 +12,7 @@ import { useRouter } from 'vue-router';
 import LabelDiv from '../widgets/LabelDiv.vue';
 import LabelSelect from '../widgets/LabelSelect.vue';
 import LabelInput from '../widgets/LabelInput.vue';
-import LabelCollapsable from '../widgets/LabelCollapsable.vue'
+import LabelCollapsable from '../widgets/LabelCollapsable.vue';
 const FLAW_BASE_URI = '/osidb/api/v1/flaws';
 // const FLAW_BASE_URI = `http://localhost:5173/tests/3ede0314-a6c5-4462-bcf3-b034a15cf106`;
 const putHandler = http.put(`${FLAW_BASE_URI}/:id`, async ({ request }) => {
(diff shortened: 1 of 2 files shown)


alizard0:
A small comment for the reviewers, I am not sure if that's the best way to prove that the element is not rendered.

https://github.com/RedHatProductSecurity/osim/pull/143/files#diff-967a205363234f686e2f6df964ed2631bc6e26c2ba4f9d69abad87898d0d3461R276

src/components/__tests__/FlawForm.spec.ts
```diff
@@ -12,6 +12,7 @@ import { useRouter } from 'vue-router';
 import LabelDiv from '../widgets/LabelDiv.vue';
 import LabelSelect from '../widgets/LabelSelect.vue';
 import LabelInput from '../widgets/LabelInput.vue';
+import LabelCollapsable from '../widgets/LabelCollapsable.vue'
```
tboyetteredhat, line 15: We need to fix lint issue. 
Screenshot from 2024-04-04 09-37-49

tboyetteredhat (changes requested):

tboyetteredhat (approved):

JakubFrejlach (approved):
I would recommend to squash these two commits into one and rebase it as such commits which tend to address syntax/linting/etc. are not really adding any value and git history will be much more readable without them. But just my personal preference.

superbuggy:
> I would recommend to squash these two commits into one and rebase it as such commits which tend to address syntax/linting/etc. are not really adding any value and git history will be much more readable without them. But just my personal preference.

To this point, formatting commits being in dedicated commits makes it possible to ignore them in git history by using an ignore revs file as in #125

JakubFrejlach:
> > I would recommend to squash these two commits into one and rebase it as such commits which tend to address syntax/linting/etc. are not really adding any value and git history will be much more readable without them. But just my personal preference.
> 
> 
> 
> To this point, formatting commits being in dedicated commits makes it possible to ignore them in git history by using an ignore revs file as in #125 

Yeah, but this is a different scenario. First commit here introduces the linting error, second commit here fixes the linting error. Having both these commits adds no value, better approach here is to just squash/fixup the commits so the linting error is never introduced in the first place.

PR #146, opened 2024-04-05
[OSIDB-2393] Refactor IndexView and AdvancedSearch Logic

tboyetteredhat:
Move fetch logic to composable.
Use composable fetch logic for both IndexView and SearchPage


src/components/IssueSearchAdvanced.vue
```diff
@@ -10,13 +9,14 @@ type Facet = {
   value: string;
 };
 
-const emit = defineEmits<{
-  'issues:load': [any[]];
+const props = defineProps<{
+  isLoading: boolean;
 }>();
 
+const emit = defineEmits(['set:filters']);
```
superbuggy, line 16: We tend to use the type signature declaration style for emits to leverage typescript checks, declaring a parameter type for `set:filters` would be useful here.

src/composables/useFlaws.ts
```diff
@@ -0,0 +1,64 @@
...
+  return {
+    issues,
+    isFinalPageFetched,
+    isLoading,
+    offset,
+    pagesize,
+    loadFlaws,
+    loadMoreFlaws
+  };
+
```
superbuggy, line 63: This is a good move, and completes the consolidation that began with reusing IssueQueue.

superbuggy: Do you think we need to expand any test coverage to give us some extra certainty while refactoring this or do you feel like we are covered?

src/composables/useFlaws.ts
```diff
@@ -0,0 +1,64 @@
...
+    isLoading.value = true;
+    issues.value = [];
+
+    getFlaws(offset.value, 100, params.value)
+      .then((response) => {
+        if (response.data.results.length < pagesize) {
+          isFinalPageFetched.value = true;
+        }
+        issues.value = response.data.results;
+        offset.value += response.data.results.length; // Increase the offset for next fetch
```
superbuggy, line 23: `response.data` has some properties that handle pagination offset values as well, I forget, can we use that to determine the offset? I guess there may not be a functional difference, if that's the case this is fine. I seem to recall this working well.

superbuggy (changes requested):
Excellent work, small changes requested

src/composables/useFlaws.ts
tboyetteredhat: response.data has only `count` property for all issue count.

src/composables/useFlaws.ts
tboyetteredhat: Thanks. I will add test for that.

src/composables/useFlaws.ts
JakubFrejlach: There should be two additional properties, `previous` and `next` which basically includes crafted URLS with offset already applied based on limit. When `next` is empty, you are on the last page.

tboyetteredhat: Thanks for your information. @JakubFrejlach  I will use next property for offset.

src/views/FlawSearchView.vue
```diff
@@ -1,67 +1,82 @@
...
+  };
+}
 
-function setIssues(loadedIssues: any[]) {
-  issues.value = loadedIssues;
+function setTableFilters(newFilters: any = {}) {
+  tableFilters.value = {
+    ...newFilters.value
+  };
 }
```
JakubFrejlach, line 61: What is the difference between the filter and table filter here? :thinking:

JakubFrejlach (approved):
One question, otherwise looking good.

src/views/FlawSearchView.vue
tboyetteredhat: `filter` is from AdvancedSearch filter.
`tableFilter` is myIssue and active issue filter on the table.

image

src/components/__tests__/FlawSearchView.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+}));
+
+vi.mock('jwt-decode', () => ({
+  default: vi.fn(() => ({
+    sub: '1234567890',
+    name: 'Test User',
+    exp: Math.floor(Date.now() / 1000) + 60 * 60 * 24 * 365,
+  })),
+}));
+
```
superbuggy, line 28: Thanks for taking care of this warning

src/components/__tests__/FlawSearchView.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+vi.mock('../../composables/useFlaws', () => ({
+  useFlaws: vi.fn(() => ({
+    issues: [],
+    isLoading: true,
+    isFinalPageFetched: false,
+    loadFlaws: vi.fn(),
+    loadMoreFlaws: vi.fn(),
+  })),
+}));
+
```
superbuggy, line 48: I agree this is probably the right approach

src/components/__tests__/IssueSearchAdvanced.spec.ts
```diff
@@ -32,4 +32,32 @@ describe('IssueSearchAdvanced', () => {
...
+    await selectDropdown.setValue(selectDropdown.findAll('option')[1].element.value);
+    await selectDropdown.trigger('change');
+    const inputField = wrapper.find('input.form-control');
+    await inputField.setValue('test'); 
+    const searchButton = wrapper.find('button[type="submit"]');
+    expect(searchButton.exists()).toBeTruthy();
+    await searchButton.trigger('submit');
+    await flushPromises();
+    const setFilterEvent = wrapper.emitted('set:filters');
+    expect(setFilterEvent[0][0]).toEqual({ acknowledgments__name: 'test' });
```
superbuggy, line 61: I like the thoroughness, but ultimately we may want to test the external behavior of the network request, so we're not testing internal implementation. For example, if the `set:filters` event is renamed this test will break although the functionality could be correct.

Instead we might want to spy/intercept this network request here

https://github.com/RedHatProductSecurity/osim/blob/b38a6274cddf3fb85a1bbffec320bfed6a9a039e/src/components/IssueSearchAdvanced.vue#L136

This spy setup might look something like

```ts
import { http, HttpResponse } from 'msw';
import { setupServer } from 'msw/node';


const getParamsHandler = http.get('/osidb/api/v1/flaws', async ({ request }) => {
  const reader = request.params?.getReader();

  if (!reader) {
    throw new Error('Request body is not available');
  }

  const result = await reader.read();

  if (result.done) {
    throw new Error('Request body stream is already read to the end');
  }

  const decoder = new TextDecoder('utf-8');
  const paramsDecoded = decoder.decode(result.value);
  return paramsDecoded;
});

const server = setupServer(getParamsHandler);

```
and include setup and teardown:
```ts
beforeAll(() => {
  server.listen({ onUnhandledRequest: 'error' });
}
// ...
afterEach(() => server.resetHandlers());
// ...
afterAll(() => server.close());
```
Then you'd make a function for the mocked request...
```ts
function mockedAdvancedSearch(params: Record<string, string>) {
  return axios({
    method: 'get',
    url: '/osidb/api/v1/flaws',
    params,
  })
    .then((response) => {
      // Not sure about this part
    })
    .catch(console.error);
}
```

```ts
// .. in test 'should emit with filters' for IssueSearchAdvanced.spec.ts
    const result = await mockedPutFlaw(flaw.uuid, flaw);
    expect(result.owner).toBe('networking test owner');
```

superbuggy: I haven't run any of the code I've written here but it is adapted from https://github.com/RedHatProductSecurity/osim/blob/c46130106f74763ecbd8aa7314d82934d38f419f/src/components/__tests__/FlawForm.spec.ts

src/components/__tests__/IssueSearchAdvanced.spec.ts
```diff
@@ -32,4 +32,32 @@ describe('IssueSearchAdvanced', () => {
...
+  it('should emit on search button click', async () => {
+    const searchButton = wrapper.find('button[type="submit"]');
+    expect(searchButton.exists()).toBeTruthy();
+    await searchButton.trigger('submit');
+    await flushPromises();
+    const setFilterEvent = wrapper.emitted('set:filters');
+    expect(setFilterEvent[0][0]).toEqual({});
+  });
+
+  it('should emit with filters on search button click', async () => {
```
superbuggy, line 50: If you end up changing the test below, this description should change to reflect the nature of the test

src/components/__tests__/IssueSearchAdvanced.spec.ts
tboyetteredhat: I think since we are writing test for useFlaws and IssueSearchAdvanced. we don't need to mock api. just mock function in useFlaws. So IssueSearchAdvanced.spec.ts file, we just test that emit is called or not.
In useFlaws.spec.ts, we are using getFlaws from flawservice. i think we can mock that getFlaws instead of mocking api. what do you think? @superbuggy

superbuggy: Ah yeah that makes way more sense

superbuggy (approved):

PR #150, opened 2024-04-08
Revises Flaw validations

superbuggy:
- CVE validation on OSIM now matches OSIDB
- New error messages
- CVEs can be blank (OSIDB seems to complain for me still though)
- Reported Date validation has been added (doesn't show on initial load for Create Flaw for some reason)
- Comment#0 minimum length is back to one from 1

Handles:
- OSIDB-2260
- OSIDB-2453
- OSIDB-2311

Changes comment#0 error messages to indicate `Comment#0` instead of `description` and remove arbitrary 10 character length.

### DIFF
--- a/src/composables/useFlawModel.ts
+++ b/src/composables/useFlawModel.ts
@@ -47,7 +47,14 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), onSaveSuccess: 
     if (!validatedFlaw.success) {
       return;
     }
-    postFlaw(validatedFlaw.data)
+    // Remove any empty fields before request
+    const flawForPost: any = Object.assign({}, validatedFlaw.data);
+    Object.entries(flawForPost).forEach(([key, value])=>{
+      if (value === '') {
+        delete flawForPost[key];
+      }
+    });
+    postFlaw(flawForPost)
       .then(createSuccessHandler({ title: 'Success!', body: 'Flaw created' }))
       .then((response: any) => {
         router.push({
--- a/src/composables/useFlawModel.ts
+++ b/src/composables/useFlawModel.ts
@@ -86,6 +86,7 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), onSaveSuccess: 
   }
 
   async function updateFlaw() {
+    isSaving.value = true;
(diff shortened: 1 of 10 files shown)


src/composables/useFlawModel.ts
```diff
@@ -47,7 +47,14 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), onSaveSuccess:
...
       return;
     }
-    postFlaw(validatedFlaw.data)
+    // Remove any empty fields before request
+    const flawForPost: any = Object.assign({}, validatedFlaw.data);
+    Object.entries(flawForPost).forEach(([key, value])=>{
+      if (value === '') {
+        delete flawForPost[key];
+      }
+    });
```
d10n, line 56: This is fine because it's iterating over Object.entries instead of flawForPost, but I might weakly, stylistically prefer this instead of `delete`
```suggestion
const flawForPost: any = Object.fromEntries(Object.entries(validatedFlaw.data).filter(([key, value]) => value !== ''))
```
(don't commit from this suggestion directly)

Use your preference, just my thought

src/utils/helpers.ts
```diff
@@ -79,4 +79,4 @@ export const deepMap = (transform: (arg: any) => any, object: DeepMappable): any
     , object
   );
 
-export const cveRegex = /^\s*(CVE-\d{4}-\d{4,7})\s*$/;
+export const cveRegex = /^CVE-(?:1999|2\d{3})-(?!0{4})(?:0\d{3}|[1-9]\d{3,})$/;
```
d10n, line 82: Used in Navbar too, check that changing semantics (removal of \s*) is fine

src/composables/useFlawModel.ts
superbuggy: That is much deliciouser. 

TBH, this was 🍳🧠 code. Even with Object.{assign,entries}, I think for non-primitives, references rather than values are copied so `delete` is potentially hazardous (though less in this case since it's mutating validatedFlaw, which actually could have entangled references with the `flaw`)

d10n (approved):

tboyetteredhat (approved):

PR #151, opened 2024-04-09
Fix/osidb 2454 attribution lists bugs

C-Valen:
Applies changes of OSIDB-2320:
- :bug: Expands `editableList` on new draft item added
- :bug: Prevents form submitting on `References` and `Acknowledgements` draft removals
- :bug: Fixes index errors on draft attribution removals
- :bug: Fixes `Affiliation` label text overflow on `Acknowledgements` items

### DIFF
--- a/src/components/IssueFieldAcknowledgments.vue
+++ b/src/components/IssueFieldAcknowledgments.vue
@@ -76,7 +76,7 @@ function handleDelete(uuid: string, closeModal: () => void) {
           </div>
           <button
             class="btn osim-cancel-new-acknowledgment"
-            @click="emit('acknowledgment:cancel-new', items[itemIndex])"
+            @click.prevent="emit('acknowledgment:cancel-new', items[itemIndex])"
           >
             <i class="bi bi-x" />
           </button>
--- a/src/components/IssueFieldAcknowledgments.vue
+++ b/src/components/IssueFieldAcknowledgments.vue
@@ -75,8 +75,9 @@ function handleDelete(uuid: string, closeModal: () => void) {
             />
           </div>
           <button
+            type="button"
             class="btn osim-cancel-new-acknowledgment"
-            @click.prevent="emit('acknowledgment:cancel-new', items[itemIndex])"
+            @click="emit('acknowledgment:cancel-new', items[itemIndex])"
           >
             <i class="bi bi-x" />
           </button>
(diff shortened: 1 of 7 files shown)


src/components/IssueFieldAcknowledgments.vue
```diff
@@ -76,7 +76,7 @@ function handleDelete(uuid: string, closeModal: () => void) {
           </div>
           <button
             class="btn osim-cancel-new-acknowledgment"
-            @click="emit('acknowledgment:cancel-new', items[itemIndex])"
+            @click.prevent="emit('acknowledgment:cancel-new', items[itemIndex])"
```
superbuggy, line 79: can also add `type="button"` here if this results in any weird side-effects

src/components/widgets/EditableList.vue
```diff
@@ -165,7 +177,7 @@ function commitEdit(index: number) {
       >
         Save Changes to {{ entityNamePlural }}
       </button>
-      <button type="button" class="btn btn-secondary" @click.prevent="emit('item:new')">
+      <button type="button" class="btn btn-secondary" @click.prevent="addItem()">
```
superbuggy, line 180: if `type="button"` is present, does that cover the use case for `.prevent`?

src/composables/useFlawAttributionsModel.ts
```diff
@@ -97,7 +97,7 @@ export function useFlawAttributionsModel(flaw: Ref<ZodFlawType>, onSaveSuccess:
   }
 
   function cancelAddAcknowledgment(acknowledgment: ZodFlawAcknowledgmentType) {
-    flawAcknowledgments.value = flawAcknowledgments.value.filter((r) => r !== acknowledgment);
+    flawAcknowledgments.value.splice(flawAcknowledgments.value.indexOf(acknowledgment),1);
   }
```
superbuggy, line 102: Ah was that the source of the bug with indexes?

superbuggy (commented):
Couple small changes suggested, other than that this looks good

src/components/IssueFieldAcknowledgments.vue
C-Valen: Thanks for the review @superbuggy 
Indeed using type button also works properly. I changed all the prevents on click to use type button instead.

src/components/widgets/EditableList.vue
C-Valen: Yeap, it does, and has been corrected to use "type button" instead. 💯

tboyetteredhat (approved):
LGTM

PR #154, opened 2024-04-10
Update features to update public date of embargoed flaw

qduanmu:


features/utils.py
```diff
@@ -113,15 +121,16 @@ def go_to_home_page(browser):
     home_page.flaw_list_exist()
 
 
-def go_to_specific_flaw_detail_page(browser, cve_id=TEST_FLAW_CVE_ID):
+def go_to_specific_flaw_detail_page(browser, embargoed=False):
     """
     Enter first flaw detail page which displayed in index page.
     """
+    cve_id = EMBARGOED_FLAW_CVE_ID if embargoed else PUBLIC_FLAW_CVE_ID
```
huiwangredhat, line 128: I think we need more than two cve_id to do test, so it's better to keep the cve_id pamas.

features/utils.py
```diff
@@ -113,15 +121,16 @@ def go_to_home_page(browser):
     home_page.flaw_list_exist()
 
 
-def go_to_specific_flaw_detail_page(browser, cve_id=TEST_FLAW_CVE_ID):
+def go_to_specific_flaw_detail_page(browser, embargoed=False):
```
huiwangredhat, line 124: The other file calls this function, e.g., Line 118 of flaw_filter.py, that should be updated together.

huiwangredhat (approved):
LGTM.

PR #156, opened 2024-04-10
📝 Contribution documents

C-Valen:
## Checklist:

- [X] Linting passed
- [X] Type checks passed
- [X] Tests suite passed
- [x] Commits consolidated
- [ ] Changelog updated (n/a)
- [ ] Test cases added/updated (n/a)
- [ ] Jira ticket updated (n/a)

## Summary:

Provides documents for pull request description template and contribution guidelines.

## Changes:

- Adds `pull_request_template.md` to  `.github` dir
- Adds `contribution.md` to `/docs`

## Considerations:

- Just a draft proposal for both documents, we can use this PR to discuss/adapt the info provided or huddle about it
- I think we should also update [OSIM Developer Guide](https://github.com/RedHatProductSecurity/osim/blob/doc/contribution-docs/docs/overview.md)
- Would be nice to extend [OSIM Developer Guide](https://github.com/RedHatProductSecurity/osim/blob/doc/contribution-docs/docs/overview.md) with some more information like `Ecosystem`, `Useful tools` ...

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,23 @@
+# [OSIDB-ID] [Title]
+
+## Checklist:
+
+- [ ] Linting passed
+- [ ] Type checks passed
+- [ ] Tests suite passed
+- [ ] Commits consolidated
+- [ ] Changelog updated
+- [ ] Test cases added/updated
+- [ ] Jira ticket updated
+
+## Summary:
+
+[Replace with a brief summary of the changes introduced by this PR.]
+
+## Changes:
+
+[Replace with a detailed description of the changes made in this PR, including any new features, enhancements, bug fixes, or refactorings.]
+
+## Considerations:
+
+[Replace with any additional considerations, notes, or instructions for reviewers.]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,66 @@
+# Contributing to OSIM
+
+Thank you for considering contributing to OSIM!
+
+Your contributions help improve the project for everyone.
+
+The following is a set of guidelines for contributing to OSIM project. Feel free to propose changes to this document in a pull request. 
+
+Please read the [OSIM Developer Guide](/docs/overview.md) first for general overview on the project ecosystem, practices and configuration.
+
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
+# [OSIDB-ID] [Title]
+
+## Checklist:
+
+- [ ] Linting passed
+- [ ] Type checks passed
+- [ ] Tests suite passed
```
JakubFrejlach, line 7: I am not sure about these 3 points, they seem kinda redundant to me as they are all checked via github workflow. Also making some additional changes to the PR might cause any of these to fail again and thus rendering this checkbox outdated.

Idea: On the other hand, we currently only run one github workflow which just runs the deploy script which does all the linting, type check, etc. It is sometimes kinda hard to find from one long log file what exactly failed.

C-Valen: > I am not sure about these 3 points, they seem kinda redundant to me as they are all checked via github workflow. Also making some additional changes to the PR might cause any of these to fail again and thus rendering this checkbox outdated.

That's a good consideration, being true that GH workflow performs automatic checks on those I think if we will have a PR checklist they should be considered as it kind of create the habit of checking those points before pushing/creating PR.

As I see it GH workflow is to ensure code is correct, and checklist points to make the developer conscious on the common actions before opening a PR.

> Idea: On the other hand, we currently only run one github workflow which just runs the deploy script which does all the linting, type check, etc. It is sometimes kinda hard to find from one long log file what exactly failed.

So you suggest to have an independent workflow script with only required steps for PRs? That'd be a a good idea as well.

JakubFrejlach: For example in OSIDB we have a github workflow setup like this:
image

superbuggy: I like this in theory, I have to think through how this might change our build process

d10n (approved):

.github/pull_request_template.md
d10n: For now, the checklist doesn't do any harm, but we can remove items that get automated with PR push actions

PR #158, opened 2024-04-10
OSIDB-2308: Fix CVSS composable

JakubFrejlach:
This PR:
* makes variables in CVSS composable more explicit (eg. versions, issuers)
* fixes cvss data search - version and issuer need to be both specified as with only issuer specified it might happen that CVSSv2 score was found first and used

Consideration:
CVSS composable should probably be redesigned a little bit a generalized in the future as there will be CVSSv4 as well

This partially solves OSIDB-2308 (sub-task OSIDB-2389)

### DIFF
--- a/src/composables/useCvssScoresModel.ts
+++ b/src/composables/useCvssScoresModel.ts
@@ -6,11 +6,19 @@ import type { ZodFlawType } from '@/types/zodFlaw';
 const onSucceed = createSuccessHandler({ title: 'Success!', body: 'Saved CVSS Scores' });
 const onError = createCatchHandler('Error updating Flaw CVSS data');
 
+// TODO: This composable should be ideally refactored into a more modular
+// solution when CVSSv4 starts being used
 export function useCvssScoresModel(flaw: Ref<ZodFlawType>) {
+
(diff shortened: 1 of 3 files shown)


superbuggy (approved):
I appreciate the variable naming in particular, this rocks

tboyetteredhat (approved):
Nice job 🚀

src/composables/useCvssScoresModel.ts
```diff
@@ -6,11 +6,19 @@ import type { ZodFlawType } from '@/types/zodFlaw';
...
+      (assessment) => assessment.issuer === issuer && assessment.cvss_version === version
+    );
+  }
+
   const wasCvssModified = ref(false);
-  const redHatCvssData = flaw.value.cvss_scores.find((assessment) => assessment.issuer === 'RH');
-  const flawRhCvss = ref(
-    redHatCvssData || {
+  const flawRhCvss3 = ref(
+    getCvssData('RH', 'V3') || {
```
superbuggy, line 21: Much better

superbuggy (commented):
🎉 for naming specificity

PR #176, opened 2024-04-17
Fix log file permissions in container

d10n:
# [OSIDB-ID] Fix log file permissions in container

## Checklist:

- [ ] Linting passed
- [ ] Type checks passed
- [ ] Tests suite passed
- [ ] Commits consolidated
- [ ] Changelog updated
- [ ] Jira ticket updated

## Summary:

/etc/nginx/nginx.conf is owned by root, but the entrypoint runs as the user, so the entrypoint script can't set the log paths.
Fix it by using symlink indirection.


costaconrado (approved):
LGTM.

PR #178, opened 2024-04-18
OSIDB-2419: Fix NVD diff highlight and update changelog

tboyetteredhat:
# [OSIDB-2419] [Difference between NVD and RH CVSS score not shown properly]

## Checklist:

- [X] Linting passed
- [X] Type checks passed
- [X] Tests suite passed
- [X] Commits consolidated
- [X] Changelog updated
- [X] Test cases added/updated
- [X] Jira ticket updated

## Summary:

Fixed NVD and RH CVSS score diff highlight when score values are integer

## Changes:

Implementing CVSS score formating


src/composables/useCvssScoresModel.ts
```diff
@@ -32,13 +32,15 @@ export function useCvssScoresModel(flaw: Ref<ZodFlawType>) {
 
   const flawNvdCvss3 = computed(() => getCvssData('NIST', 'V3'));
 
+  const formatScore = (score: any) => score?.toFixed(1);
```
C-Valen, line 35: Is score always gona be a number? In that case I think we could specify the type or in other cases cast to number to avoid issues on the `toFixed` method.
```suggestion
  const formatScore = (score: number) => score?.toFixed(1);
```

C-Valen (approved):
LGTM just pointed minor possible consideration.

superbuggy (approved):

PR #179, opened 2024-04-18
💄 OSIDB-2512: Show url on references links

C-Valen:
## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Test cases added/updated _(N/A)_
- [x] Jira ticket updated

## Summary:

Shows the references link as an url instead of having the link on the timestamp.
Timestamp is kept as an independent field for info purposes.

## Changes:

- Shows url on references links
- Changes the layout on info fields at references
- Improves the contrast ratio on "External" badges text

## Considerations:

- The layout change to vertical (of badge - timestamp - url)  was considered as some urls can be quite long so they should have space inline

### DIFF
--- a/src/components/IssueFieldReferences.vue
+++ b/src/components/IssueFieldReferences.vue
@@ -57,7 +57,7 @@ function handleDelete(uuid: string, closeModal: () => void) {
           >
             {{ referenceTypeLabel(type) }}
           </span>
-          <a class="ref-link" :href="url" target="_blank">{{ url }}</a>
+          <a class="m-1" :href="url" target="_blank">{{ url }}</a>
         </div>
         <div class="form-group mt-2">
           <LabelStatic v-model="item.description" label="Description" hasTopLabelStyle />
@@ -162,10 +162,6 @@ select.osim-reference-types {
 .info-group {
   display: flex;
   flex-direction: column;
-}
-
-.ref-link {
-  margin: 5px;
 }
 
 .badge {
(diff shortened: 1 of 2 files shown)


superbuggy:
I wonder if we can shorten the badge text for RHSBs and put the badge on the same line as the timestamp. The other thing we could  do to shorten urls is to just show the domain of the link and a link icon

C-Valen:
> I wonder if we can shorten the badge text for RHSBs and put the badge on the same line as the timestamp. The other thing we could do to shorten urls is to just show the domain of the link and a link icon

Good points @superbuggy!

Just noting that timestamps were finally removed as per stakeholders request (they are no needed).

About the the shorten links I thought the same at first, generally I'd take that approach, but I think for this specific scenario it's kind of useful for the users to see the full URL that the item is referencing.

On the layout, I have weak preference to have it in vertical just because the path to follow when looking into references items info looks simpler to me.

src/components/IssueFieldReferences.vue
```diff
@@ -45,19 +45,20 @@ function handleDelete(uuid: string, closeModal: () => void) {
...
-          <span class="me-2">{{ updated_dt }}</span>
+      <template #default="{ item: { url, type, ...item } }">
+        <div class="info-group">
           <span
             class="badge rounded-pill"
             :class="{
               'bg-primary': type === 'ARTICLE',
               'bg-warning': type !== 'ARTICLE',
             }"
+            :style="type !== 'ARTICLE' ? 'color: black' : ''"
```
superbuggy, line 56: I think this is fine for the time being but I'm wondering if we want that black on yellow contrast everywhere--what do you think? We can also lighten the yellow too

C-Valen: I think the other elements with similar style are already with black font over yellow bg.
For example the workflow buttons on status field:

image

No preference from me, we can discuss the style on this "warning like" elements in the future.

PR #183, opened 2024-04-23
Add Total Count for IndexView, Search Page

tboyetteredhat:
# [OSIDB-2438] [Add count of results returned]

## Checklist:
- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Add Total, Loaded Count for IssueQueue

## Changes:

Add Total Count for IssueQueue table

### DIFF
--- a/src/components/IssueQueue.vue
+++ b/src/components/IssueQueue.vue
@@ -20,6 +20,7 @@ const props = defineProps<{
   issues: any[];
   isLoading: boolean;
   isFinalPageFetched: boolean;
+  total: number
 }>();
 
 const issues = computed<any[]>(() => props.issues.map(relevantFields));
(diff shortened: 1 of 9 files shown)


C-Valen (approved):
LGTM

I'd recommend adding bold font type to make it clearer to the user.

And I'd place it within "My Issues" and "Open Issues" check-boxes, but that's just a personal preference.

image

src/components/IssueQueue.vue
```diff
@@ -157,6 +157,7 @@ watch(params, () => {
         <span class="visually-hidden">Loading...</span>
       </span>
       <span v-if="isLoading"> Loading Flaws&hellip; </span>
+      <span class="float-end fw-bold"> Total: {{ issues.length }}</span>
```
superbuggy, line 160: I think we should show the number of fetched results (`issues.length`) and the `.count` from the response from GET requests to `https://osidb-stage.prodsec.redhat.com/osidb/api/v1/flaws`

superbuggy: https://github.com/RedHatProductSecurity/osim/blob/fix/OSIDB-2390-sequential-saving-on-flaws/src/composables/useFlaws.ts#L17-L63

You can access `response.data.count` in the highlighted block above and use a `ref` to capture the number of results found by OSIDB.

superbuggy (changes requested):

src/components/IssueQueue.vue
tboyetteredhat: Thanks for your feedback. I will update to show both total and loaded count.

d10n (approved):

superbuggy (approved):

PR #186, opened 2024-04-24
🩹 OSIDB-2562 Hide source column

C-Valen:
## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [X] Jira ticket updated

## Summary:

Temporary hiding the `Source` column on Flaw queues and renaming `Source` field labels to `CVE Source`:

> The current value of the source field is where the flaw was reported. This information is not very valuable on this page. Instead, we need to track what system the flaw was created in (SNOW, Assembler, etc; ). Today, we only get flaws from assembler or a direct users. So we should hide the source column on Index page for now. When we do an integration with SNOW, we could then expose the field with more meaningful data.

> Additionally, we should rename the source field on the CVE page to 'CVE Source' so that it is more clear what the field represents. 

## Changes:

- Hidden `Source` column from `IssueQueue` and `IssueQueueItem`
- Renamed `Source` field label to `CVE Source` at `FlawForm`
- Renamed `Source` field to `CVE Source` at `AdavancedSearch`
- Updated `FlawForm` tests with new `CVE Source` label

## Considerations:

- `CVE Source` remains editable on `FlawForm`
- `CVE Source` is still a valid search field for `AdvancedSearch`
- After SNOW integration review possible new values and field behavior (may be static and auto-generated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 * Implemented read-only mode (network requests involving write operations are disabled)
 * Disables form on save
 * Implemented Advanced Search Query to URL
+* `Source` column hidden on flaw lists
+* `Source` field renamed to `CVE Source` on flaw form and advanced search
 
 ### Fixed
 * Fixed overlap on edit buttons at References and Acknowledgements items
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -220,7 +220,7 @@ const onReset = () => {
           />
           <LabelSelect
             v-model="flaw.source"
-            label="Source"
+            label="CVE Source"
             :options="flawSources"
             :error="errors.source"
           />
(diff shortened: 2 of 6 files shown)


tboyetteredhat (approved):

d10n:
Looks good, only weak request I have is: if commenting the code instead of deleting, say why, because it might not be clear when the right time to delete the comments is

d10n (approved):

superbuggy:
> Looks good, only weak request I have is: if commenting the code instead of deleting, say why, because it might not be clear when the right time to delete the comments is

I like this process--keeps my 'delete 'em all (let the gc sort em out)' tendencies in check

PR #187, opened 2024-04-26
♻️ OSIDB 1596: Unembargo confirmation on save flaw

C-Valen:
## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Test cases added/updated (_N/A_)
- [x] Jira ticket updated

## Summary:

Refactor of unembargo workflow to make the confirmation pop up shown on save flaw.

## Changes:

- Fixed stuck saving state on flaw update/creation (restore `isSaving` variable before return)
- Added a `isValid` method on `useFlawModel` that returns the boolean result of the flaw validation
- Refactored `IssueFieldEmbargo` for the "onSave" confirmation
- Added control of unembargo modal on `FlawForm` 
- Added reset button on embargo field (visible only if going to be unembargoed)
- Added warning tooltip on embargo field (only if going to be unembargoed)
- Other minor style improvements on `IssueFieldEmbargo`
- Removed unused `committedFlaw` var at `FlawEditView

## Considerations:

- `isValid` method purpose is to be able to check the result of the flaw validation without any UI information. For this specific case it was required to show the unembargo modal only if the validation is succeeded so if there are validation errors the modal is not shown and the validation alerts are not duplicated.
- `onSubmit` at `FlawForm` now shows the unembargo confirmation modal if the flaw is gonna be unembargoed and the form is valid (if in edit mode). After a correct confirmation, the flaw is updated from an `update:flaw` emitter sent from the modal to the parent form.
- A non saved unembargoed state can either be reset with the specific reset button on embargo field or reset all on the form. The field button is just to make more obvious and easy to the users to reset the field before saving if needed.

### DIFF
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -24,7 +24,6 @@ import { useFlawModel } from '@/composables/useFlawModel';
 import { fileTracker, type TrackersFilePost } from '@/services/TrackerService';
 import type { ZodFlawType } from '@/types/zodFlaw';
 
-
 const props = defineProps<{
   flaw: any;
   mode: 'create' | 'edit';
@@ -73,6 +72,7 @@ const {
   saveAcknowledgments,
   deleteAcknowledgment,
   isSaving,
+  isValid,
   errors,
 } = useFlawModel(props.flaw, onSaveSuccess);
 
@@ -87,15 +87,21 @@ watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any 
   onReset();
 });
 
+const isPublic = computed(() => !initialFlaw.value?.embargoed);
+const showUnembargoingModal = ref(false);
+const unembargoing = computed(() => !isPublic.value && !flaw.value.embargoed);
+
 const onSubmit = async () => {
   if (props.mode === 'edit') {
-    updateFlaw();
-  }
-  if (props.mode === 'create') {
+    if(isValid() && unembargoing.value) {
+      showUnembargoingModal.value = true;
+    } else {
+      updateFlaw();
+    }
+  } else if (props.mode === 'create') {
     createFlaw();
   }
 };
-
 
 const showSummary = ref(flaw.value.summary && flaw.value.summary.trim() !== '');
 const showStatement = ref(flaw.value.statement && flaw.value.statement.trim() !== '');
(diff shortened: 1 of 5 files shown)


src/components/FlawForm.vue
```diff
@@ -87,16 +87,22 @@ watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any
   onReset();
 });
 
+const isPublic = computed(() => !initialFlaw.value?.embargoed);
+const showUnembargoingModal = ref(false);
```
tboyetteredhat, line 91: I think we can use useModal composable. https://github.com/RedHatProductSecurity/osim/blob/main/src/composables/useModal.ts

C-Valen: Hey @tboyetteredhat thanks for the review.

The `useModal` composable is used in the `IssueFieldEmbargo` component, do you mean to substitute the property on `showUnembargoingModal` for `useModal` methods?

I think that might require to move the modal composable from the child to the parent?

tboyetteredhat: Oh I see. I think we don't need to change it. Looks good to me.

tboyetteredhat (approved):

src/components/IssueFieldEmbargo.vue
```diff
@@ -90,15 +125,58 @@ function handleConfirm() {
             </Modal>
           </div>
         </div>
+        
+
```
C-Valen, line 129: Remove unnecessary empty lines
```suggestion
```

superbuggy (approved):

src/components/IssueFieldEmbargo.vue
```diff
@@ -9,47 +9,82 @@ import LabelCheckbox from '@/components/widgets/LabelCheckbox.vue';
 const props = defineProps<{
   flawId: string | null | undefined;
   isFlawNew: boolean;
+  isPublic: boolean;
 }>();
 
 const modelValue = defineModel<boolean | undefined>();
+const showModal = defineModel<boolean | undefined>('showModal');
 
 // const isFlawNew = computed(() => modelValue.value === undefined);
```
C-Valen, line 18: Remove duplicated var
```suggestion
```

PR #188, opened 2024-04-26
Change create flaw related test cases, use relative locators

RichardXuan:


RichardXuan:
@qduanmu Should we merge this PR or wait for the bug fix.

PR #210, opened 2024-05-16
Add test case for recover affect from deleting

huiwangredhat:
# [OSIDB-1699] [Affect-recover]

### DIFF
--- a/features/steps/flaw_detail.py
+++ b/features/steps/flaw_detail.py
@@ -4,7 +4,6 @@ from datetime import date, datetime
 from behave import when, then
 from selenium.webdriver.common.by import By
 from seleniumpagefactory.Pagefactory import ElementNotFoundException
-from features.utils import get_osidb_token
 
 from features.utils import (
     wait_for_visibility_by_locator,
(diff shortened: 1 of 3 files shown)


features/steps/flaw_detail.py
```diff
@@ -551,8 +551,10 @@ def step_impl(context):
     flaw_detail_page.check_value_exist(context.ps_component)
     context.browser.quit()
 
-@when("I delete an affect of the flaw")
-def step_impl(context):
+def click_affect_delete_btn(context):
```
RichardXuan, line 554: Better put this function into `FlawDetailPage` class

features/steps/flaw_detail.py
```diff
@@ -573,4 +581,18 @@ def step_impl(context):
...
+    context.ps_component = click_affect_delete_btn(context)
+    # Click the delete button of the affect
+    flaw_detail_page = FlawDetailPage(context.browser)
+    flaw_detail_page.click_button_with_js('affectRecoverBtn')
+
+@then("I could 'recover' the affect that I tried to delete above")
+def step_impl(context):
+    go_to_specific_flaw_detail_page(context.browser)
+    flaw_detail_page = FlawDetailPage(context.browser)
+    flaw_detail_page.check_value_exist(context.ps_component)
```
RichardXuan, line 597: Is it sufficient to only check one field value.

features/steps/flaw_detail.py
huiwangredhat: Updated.

features/steps/flaw_detail.py
huiwangredhat: Updated to check the module and component.

RichardXuan (approved):
LGTM

PR #212, opened 2024-05-20
Fix: Use object-specific updated_dt

C-Valen:
# OSIDB-2697: Error 409 updating a flaw reference

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated (_NA_)
- [x] Jira ticket updated

## Summary:

This PR aims to address an issue found by @JakubFrejlach where OSIM was sending the `updated_dt` of the flaw, to sub-models (i.e. Acks, Refs, Comments, Affects, etc.), making OSIDB respond with 409 Error "Save operation based on an outdated model instance".

## Changes:

- Renamed `getFlawUpdatedDt` function to `getUpdatedDt` on `FlawService.ts`
- Use the `options.url` on `beforeFetch` to get the specific object `update_dt`
- Removed potentially deprecated

## Considerations:

- This issue was only manifested when OSIDB's BZ Sync was disconnected, in other case the the updated_dt was not compared from OSIDB data but BZ.
- @superbuggy I removed some code that looked deprecated after this changes. Would you mind to have a look on that to ensure nothing essential was removed?

### DIFF
--- a/src/services/FlawService.ts
+++ b/src/services/FlawService.ts
@@ -42,10 +42,9 @@ export async function beforeFetch(options: OsidbFetchOptions) {
       return;
     }
     try {
-      const flaw = await getFlawUpdatedDt(flawId);
-      options.data.updated_dt = flaw.updated_dt;
-      if (!flaw.updated_dt) {
-        console.error('During multi-stage operation, an updated_dt could not be fetched', flaw, flawId);
+      const updated_dt = await getUpdatedDt(options.url);
+      if (!updated_dt) {
+        console.error('During multi-stage operation, an updated_dt could not be fetched');
         throw new Error('During multi-stage operation, an updated_dt could not be fetched');
       }
     } catch (error) {
@@ -96,10 +95,10 @@ export async function getFlaw(uuid: string): Promise<ZodFlawType> {
   }).then((response) => response.data);
 }
 
-export async function getFlawUpdatedDt(uuid: string): Promise<ZodFlawType> {
+export async function getUpdatedDt(url: string): Promise<ZodFlawType> {
   return osidbFetch({
     method: 'get',
-    url: `/osidb/api/v1/flaws/${uuid}`,
+    url: url,
     params: {
       include_fields: 'updated_dt',
     },
--- a/src/services/FlawService.ts
+++ b/src/services/FlawService.ts
@@ -8,39 +8,9 @@ import {
   type OsidbFetchOptions
 } from '@/services/OsidbAuthService';
 import { createCatchHandler, createSuccessHandler } from '@/composables/service-helpers';
-import { isFlawIdentifierValid } from '@/utils/helpers';
-
-import { ZodFlawSchema } from '@/types/zodFlaw';
-
-
-function isFlaw(data: any) {
-  return !!ZodFlawSchema.safeParse(data).success;
-}
-
-function parseFlawId(options: any) {
-  if (isFlaw(options.data) && options.data.uuid) {
-    return options.data.uuid;
-  }
-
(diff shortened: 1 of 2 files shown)


superbuggy:
It may be safe to remove those functions, as long at the `updated_dt` fetching behavior doesn't happen on inappropriate requests, it's probably fine--in retrospect, I'm not totally certain why I added as many guard clauses as I did but it was probably for data integrity fault tolerance.

C-Valen:
> It may be safe to remove those functions, as long at the `updated_dt` fetching behavior doesn't happen on inappropriate requests, it's probably fine--in retrospect, I'm not totally certain why I added as many guard clauses as I did but it was probably for data integrity fault tolerance (if data was weird on stage, the feature would still work without throwing confusing errors resulting from expected data/fields being blank/empty/falsey).

Thanks for the insights @superbuggy From the tests I performed with stage data I didn't find any broken scenario.
If there isn't any specific change you expect to made could you approve it?

src/services/FlawService.ts
```diff
@@ -42,10 +42,9 @@ export async function beforeFetch(options: OsidbFetchOptions) {
       return;
     }
     try {
-      const flaw = await getFlawUpdatedDt(flawId);
-      options.data.updated_dt = flaw.updated_dt;
-      if (!flaw.updated_dt) {
-        console.error('During multi-stage operation, an updated_dt could not be fetched', flaw, flawId);
+      const updated_dt = await getUpdatedDt(options.url);
```
JakubFrejlach, line 15: Since we no longer fetch only Flaw but the model fetched here might be various (Affect, Reference, etc.) we should change wording of the error bellow (lines 51,52)

```
console.error('Problem fetching flaw for sequential update:', error);
throw new Error('Problem fetching flaw for sequential update');
```

src/services/FlawService.ts
```diff
@@ -96,10 +95,10 @@ export async function getFlaw(uuid: string): Promise<ZodFlawType> {
   }).then((response) => response.data);
 }
 
-export async function getFlawUpdatedDt(uuid: string): Promise<ZodFlawType> {
+export async function getUpdatedDt(url: string): Promise<ZodFlawType> {
```
JakubFrejlach, line 68: Return type of this function will now not always be `Promise<ZodFlawType>` I think.

JakubFrejlach (changes requested):
Couple of smaller things, otherwise LGTM!

superbuggy (approved):

C-Valen:
> Couple of smaller things, otherwise LGTM!

Very good catches @JakubFrejlach thanks for mentioning them.
Just amended them on last commits. Please give it another look when you can.

src/services/FlawService.ts
```diff
@@ -8,49 +8,17 @@ import {
...
 
 export async function beforeFetch(options: OsidbFetchOptions) {
   if (options.data && ['PUT', 'POST'].includes(options.method.toUpperCase())) {
-    const flawId = parseFlawId(options);
-    if (flawId === null) {
-      return;
-    }
     try {
-      const flaw = await getFlawUpdatedDt(flawId);
-      options.data.updated_dt = flaw.updated_dt;
```
JakubFrejlach, line 46: One last question. When this got deleted, how does the `updated_dt` get sent in the request body?

tboyetteredhat (approved):

src/services/FlawService.ts
C-Valen: When that got deleted made the `updated_dt` being send in the request the same as before, which is wrong.
Indeed the just fetched `updated_dt` is the one that should go in the request.

Again, thanks for noting it, that was essential.

JakubFrejlach (approved):
LGTM now!

PR #222, opened 2024-05-23
Feat: Support references and acknowledgements on flaw creation

C-Valen:
# OSIDB-2319: Ensure that references / acknowledgements are sent with flaw creation

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Add support for references and acknowledgements on flaw creation, fix minor bugs, add draft flaw state.

## Changes:

- Hide save refs/acks button on flaw creation form
- Handle saving refs/acks on flaw creation if present
- Create a store to handle draft flaw data when it is not saved correctly (mainly for concatenated service operations)
- Use the draftFlaw store to keep non saved attributions on the form (as draft)
- Fix some stuck states scenarios on saving flaw/attributions unsuccessfully
- Fix wrong updated_dt initialization on blank attributions
- Fix a bug where updated_dt is fetched for attributions creation

## Considerations:

- References and ackowledgements are saved on flaw creation within the full form submit button
- Flaw is created first, uuid is taken from response and used in the refs/acks to save 
- DraftFlaw store aims to cover the issue where user's draft data on the form is lost. This scenario appears cause of the implementation of sequenced service operations, such as saving/creating multiple affects or attributions.
For the scope in this PR, when creating a flaw with multiple attributions, if one of them fails, the non saved ones will be still on form of the new EditFlaw view.

### DIFF
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -79,7 +79,7 @@ const { draftFlaw } = useDraftFlawStore();
 
 onMounted(() => {
   initialFlaw.value = deepCopyFromRaw(props.flaw) as ZodFlawType;
-  if(draftFlaw) {
+  if (draftFlaw) {
     flaw.value = useDraftFlawStore().addDraftFields(flaw.value);
     useDraftFlawStore().$reset();
   }
@@ -87,7 +87,6 @@ onMounted(() => {
 
 watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any change (though that shouldn't happen)
   initialFlaw.value = deepCopyFromRaw(props.flaw) as ZodFlawType;
-  console.log(1,useDraftFlawStore().draftFlaw);
   onReset();
 });
 
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -383,6 +383,7 @@ const toggleMitigation = () => {
           <IssueFieldReferences
             v-model="flawReferences"
             class="w-100 my-3"
+            :mode="mode"
             :error="errors.references"
             @reference:update="saveReferences"
             @reference:new="addBlankReference(flaw.embargoed)"
@@ -392,6 +393,7 @@ const toggleMitigation = () => {
           <IssueFieldAcknowledgments
             v-model="flawAcknowledgments"
             class="w-100 my-3"
+            :mode="mode"
             :error="errors.acknowledgments"
             @acknowledgment:update="saveAcknowledgments"
             @acknowledgment:new="addBlankAcknowledgment(flaw.embargoed)"
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -23,6 +23,7 @@ import CvssCalculator from '@/components/CvssCalculator.vue';
 import { useFlawModel } from '@/composables/useFlawModel';
 import { fileTracker, type TrackersFilePost } from '@/services/TrackerService';
 import { type ZodFlawType, summaryRequiredStates } from '@/types/zodFlaw';
+import { useDraftFlawStore } from '@/stores/DraftFlawStore';
 
 const props = defineProps<{
   flaw: any;
@@ -74,13 +75,19 @@ const {
 } = useFlawModel(props.flaw, onSaveSuccess);
 
 const initialFlaw = ref<ZodFlawType>();
+const { draftFlaw } = useDraftFlawStore();
 
 onMounted(() => {
   initialFlaw.value = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  if(draftFlaw) {
+    flaw.value = useDraftFlawStore().addDraftFields(flaw.value);
+    useDraftFlawStore().$reset();
+  }
 });
 
 watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any change (though that shouldn't happen)
   initialFlaw.value = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  console.log(1,useDraftFlawStore().draftFlaw);
   onReset();
 });
 
--- a/src/composables/useFlawModel.ts
+++ b/src/composables/useFlawModel.ts
@@ -26,6 +26,7 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), onSaveSuccess: 
   const flaw = ref<ZodFlawType>(forFlaw);
   const cvssScoresModel = useFlawCvssScores(flaw);
   const flawAffectsModel = useFlawAffectsModel(flaw);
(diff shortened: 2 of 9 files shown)


src/composables/useFlawModel.ts
```diff
@@ -62,15 +65,28 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), onSaveSuccess:
...
         });
-        flaw.value.uuid = response.uuid;
       })
-      .catch(createCatchHandler('Error creating Flaw'));
-
+      .catch(createCatchHandler('Error creating Flaw'))
+      .finally(() => isSaving.value = false);
     if (wasCvssModified.value) {
       await saveCvssScores();
     }
```
superbuggy, line 92: I would consider only saving acknowledgments/references if any have been added, and putting it in a separate step from the initial flaw save

src/composables/useFlawModel.ts
```diff
@@ -26,11 +27,13 @@ export function useFlawModel(forFlaw: ZodFlawType = blankFlaw(), onSaveSuccess:
   const flaw = ref<ZodFlawType>(forFlaw);
   const cvssScoresModel = useFlawCvssScores(flaw);
   const flawAffectsModel = useFlawAffectsModel(flaw);
+  const flawAttributionsModel = useFlawAttributionsModel(flaw, isSaving, afterSaveSuccess);
   const { wasCvssModified, saveCvssScores } = cvssScoresModel;
   const { affectsToSave, saveAffects, deleteAffects, affectsToDelete } = flawAffectsModel;
 
   const router = useRouter();
   const committedFlaw = ref<ZodFlawType | null>(null);
+  const { saveDraftFlaw } = useDraftFlawStore();
```
superbuggy, line 36: Is the intention for the store to be read from at some point?

superbuggy (changes requested):

src/composables/useFlawModel.ts
C-Valen: They are being saved only if any was added. That's handled by lines 73 & 76:
` if (flaw.value.acknowledgments.length > 0) ` ... `if (flaw.value.references.length > 0)` 

As it is flaw creation it is not needed to consider existent attributions. If there is any value will be a new one that needs to be saved.

Is that what you were referring to?

src/composables/useFlawModel.ts
C-Valen: Actually `draftFlaw` is being read currently in different places. May be we can remove that and just use the variable assignation.

src/composables/useFlawModel.ts
superbuggy: Here is what I was thinking...

```js
    const response = await postFlaw(flawForPost)
      .then(createSuccessHandler({ title: 'Success!', body: 'Flaw created' }))
      .then(async (response: any) => {
        router.push({
          name: 'flaw-details',
          params: { id: response?.cve_id || response?.uuid },
        });
        // We could also update the flaw with the response here, this would impact the refresh logic so as not to rely on the router for refreshing the flaw, which could be a better approach
      })
      .catch(createCatchHandler('Error creating Flaw'))
      .finally(() => isSaving.value = false);

    // This is the separate step I was referring to earlier
    if (response.uuid) {
      flaw.value.uuid = response.uuid;
      saveDraftFlaw(flaw.value);

      if (flaw.value.acknowledgments.length > 0) {
        await flawAttributionsModel.saveAcknowledgments(flaw.value.acknowledgments)
        .catch(createCatchHandler('Error creating Flaw acknowledgments'));
      }
      if (flaw.value.references.length > 0) {
        await flawAttributionsModel.saveReferences(flaw.value.references)
        .catch(createCatchHandler('Error creating Flaw references'));
      }
    }

    if (wasCvssModified.value) {
      await saveCvssScores();
    }
    isSaving.value = false;
```

C-Valen: I get your point on separating the logic but I don't see what practical benefits we get from that.
IMO refreshing the flaw from the response will complicate more the logic, than just doing the redirect from `flaw-create` to `flaw-details` using the router.

src/composables/useFlawModel.ts
C-Valen: The thing with the store is to get the draft state of the flaw (not saved values) across different components in the application. This is needed cover the problem in bulk actions (like saving multiple references when a flaw is being created) where if one fails in the middle, the rest of draft references from the user are lost.
The options I considered were:

1. Using emitters to get the state in those components
2. Have a store where components can get the draft state

The store looked less complicated to me and reusable in case we need/want to incorporate more flaw fields.

superbuggy: Okay, I think this makes sense--do you think we should migrate some of the editing functions over to use this new store? Seems like we should (not in this PR but generally).

src/stores/DraftFlawStore.ts
```diff
@@ -0,0 +1,50 @@
...
+    return fetchedFlaw;
+  }
+
+  const mergeRefs =
+    (a: ZodFlawReferenceType[], b: ZodFlawReferenceType[]) => unionWith(refsComparator, a, b);
+
+  const refsComparator = (a: ZodFlawReferenceType, b: ZodFlawReferenceType): boolean => {
+    return a.url === b.url && a.description === b.description;
+  };
+
```
superbuggy, line 31:
```suggestion
// If an element exists in both lists, the first element from the first list will be used.
```

superbuggy (approved):
Nice work!! Just suggestion to add a comment to describe the behavior of `Ramda.unionWith`

d10n (approved):
Looks great

src/components/FlawForm.vue
```diff
@@ -74,13 +75,19 @@ const {
...
   initialFlaw.value = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  if(draftFlaw) {
+    flaw.value = useDraftFlawStore().addDraftFields(flaw.value);
+    useDraftFlawStore().$reset();
+  }
 });
 
 watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any change (though that shouldn't happen)
   initialFlaw.value = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  console.log(1,useDraftFlawStore().draftFlaw);
```
JakubFrejlach: Leftover console.log?

JakubFrejlach (approved):
LGTM, one nitpick :smile:

src/components/FlawForm.vue
C-Valen: Oh that was a catch from David yesterday I supposedly removed already

JakubFrejlach: Ah, sometimes going commit by commit is not ideal :smile:

PR #227, opened 2024-05-27
Fix: Unable to save new cvss scores

C-Valen:
# Fix: Don't fetch updated_dt on object creation

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated (_NA_)

## Summary:

Fixes a bug where cvss scores couldn't be saved when editing existent flaws.
This was causes by attempting to 

## Changes:

- Removes `beforeFetch` action on `postFlawCvssScores` as if there is no cvss scores (first cvss score creation) fetch would fail.
- [Edit] Removes `beforeFetch` action on `postFlawAcknowledgment` and `postFlawReference`
- [Edit] Removes `beforeFetch` action on `postAffectCvssScore`

## Considerations:

- `beforeFetch` gets the `updated_dt` from the object to be updated, so it shouldn't be used when the object is being created for first time (e.g. Flaw Creations) cause there won't be any `updated_dt` to get from OSIDB side and action would fail.

### DIFF
--- a/src/services/FlawService.ts
+++ b/src/services/FlawService.ts
@@ -230,7 +230,6 @@ export async function advancedSearchFlaws(params: Record<string, string>) {
 
 export async function postFlaw(requestBody: any) {
   // {
-  //   "type": "VULNERABILITY",
   //   "cve_id": "string",
   //   "impact": "LOW",
   //   "component": "string",
--- a/src/services/FlawService.ts
+++ b/src/services/FlawService.ts
@@ -10,7 +10,7 @@ import {
 import { createCatchHandler, createSuccessHandler } from '@/composables/service-helpers';
 
 export async function beforeFetch(options: OsidbFetchOptions) {
-  if (options.data && ['PUT', 'POST'].includes(options.method.toUpperCase())) {
+  if (options.data && ['PUT'].includes(options.method.toUpperCase())) {
     try {
       const updated_dt = await getUpdatedDt(options.url);
       options.data.updated_dt = updated_dt;
--- a/src/services/FlawService.ts
+++ b/src/services/FlawService.ts
@@ -128,7 +128,7 @@ export async function postFlawCvssScores(flawId: string, cvssScoreObject: unknow
     method: 'post',
     url: `/osidb/api/v1/flaws/${flawId}/cvss_scores`,
     data: postObject,
-  }, { beforeFetch })
+  })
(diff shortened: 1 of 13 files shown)


src/services/FlawService.ts
```diff
@@ -128,7 +128,7 @@ export async function postFlawCvssScores(flawId: string, cvssScoreObject: unknow
     method: 'post',
     url: `/osidb/api/v1/flaws/${flawId}/cvss_scores`,
     data: postObject,
-  }, { beforeFetch })
+  })
```
superbuggy, line 131: I think this bug was caused by the changes to the flaw Id fetching--are we certain that removing this doesn't lose functionality we need in order to avoid the error related to outdated `updated_dt` that the original `beforeFetch` addressed>

JakubFrejlach: Just thinking out loud. Does it make sense to check for `updated_dt` on POSTs? You are creating the resource and thus there is not such  `updated_dt` which you could compare to determine the freshness of data. Maybe the parent resource `updated_dt` should be checked in this case? That would at least make sense for prod/stage where Bugzilla sync is turned on and every child resource is in sync with the parent on `updated_dt`, for UAT, this is however not a case

C-Valen: Yes, this bug appeared after the changes on how OSIM handles `updated_dt`.

Before was being sent the flaw's `updated_dt`, and now the `updated_dt` is fetched from OSIDB for the specific object of the operation and added to the operation's request.

The bug comes for that fetch attempt of an `updated_dt` for an object that doesn't exists.
As this is a POST method used exclusively for `cvss_scores` creation the `updated_dt` is not needed (OSIDB doesn't expects/handles it)

Apart from the theory I have tested around I didn't found any lost feature or new bugs.

C-Valen: > Is there or would there ever be a situation where the flaw's updated_dt becomes out of date in between loading the flaw and saving the CVSS score?
I think that scenario is the same as two users are in the same flaw and one edits the CVSS before the other, the second one will get outdated model instance. But I see it as the normal behavior on the updated_dt purprose.

d10n (approved):

superbuggy (approved):

PR #231, opened 2024-05-28
🐛 Don't save unmodified affects

superbuggy:
* Add deep equality check to prevent unintentionally tracking affects as modified.

# [OSIDB-2754] [OSIM: Unmodified affects are saved]

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Affects were being tracked as modified when they weren't, this fixes that issue.

## Changes:

Added a deep equality check to properly check for modified affects.

## Considerations:

Affects began triggering the watcher when loaded, which marked them as modified erroneously as soon as they loaded.

### DIFF
--- a/src/composables/useFlawAffectsModel.ts
+++ b/src/composables/useFlawAffectsModel.ts
@@ -12,6 +12,7 @@ import { useToastStore } from '@/stores/ToastStore';
 import type { ZodFlawType } from '@/types/zodFlaw';
 import type { ZodAffectType, ZodAffectCVSSType } from '@/types/zodAffect';
 import { deepCopyFromRaw } from '@/utils/helpers';
+import { equals } from 'ramda';
 
 export function useFlawAffectsModel(flaw: Ref<ZodFlawType>) {
   const wereAffectsModified = ref(false);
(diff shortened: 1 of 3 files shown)


src/composables/useFlawAffectsModel.ts
```diff
@@ -123,9 +124,14 @@ export function useFlawAffectsModel(flaw: Ref<ZodFlawType>) {
...
 
+  function hasAffectChanged(affect: ZodAffectType) {
+    const originalAffect = initialAffects.find((maybeMatch) => maybeMatch.uuid === affect.uuid);
+    return !equals(originalAffect, affect);
+  }
+
   theAffects.value.forEach((affect) => {
     watch(affect, () => {
-      if (affect.uuid) {
+      if (affect.uuid && hasAffectChanged(affect)) {
```
MrMarble, line 134: [watch](https://vuejs.org/guide/essentials/watchers.html#basic-example) callbacks have the current and previous values as parameters, is that not enough? Why do we need to use the `initialAffects` array? Just curious

```js
 watch(affect, (newAffect, oldAffect) => {
```
```suggestion
      if (affect.uuid && !equals(newAffect, oldAffect)) {
```
> github does not allow me to suggest changes on line 133 for some reason

superbuggy: That's a great question! It frustratingly does not work that way with properties of reactives :(
From https://vuejs.org/guide/essentials/watchers.html#deep-watchers
```js
const obj = reactive({ count: 0 })

watch(obj, (newValue, oldValue) => {
  // fires on nested property mutations
  // Note: `newValue` will be equal to `oldValue` here
  // because they both point to the same object!
})

obj.count++
```

MrMarble: Fair point there, but if we only need to use the `uuid` of the affect, shouldn't it be easier just to watch the uuid?
That way the `{ deep: true }` can be removed
```js
theAffects.value.forEach((affect) => {
  watch(()=>affect.uuid, () => {
    reportAffectAsModified(uuid);
  })
})

d10n (approved):

MrMarble (approved):

tboyetteredhat (approved):

C-Valen (approved):

PR #239, opened 2024-05-31
Feature: Button to expand all affects

C-Valen:
# OSIDB-2817: Button to expand all affects
## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

 Adds a new button that conditionally renders to allow the users expand all affect modules

## Changes:

- Adds`areAllModulesCollapsed` computed property on `AffectedOfferings`.
- Adds `expandAll` function on `AffectedOfferings`.
- Adds a `Expand All` new button  on the affects section

## Considerations:

- Was confirmed that only Modules (and not component details) should be expanded
- Credits to @superbuggy as this PR is just a rework from 'CollapseAll' functionality

### DIFF
--- a/src/components/AffectedOfferings.vue
+++ b/src/components/AffectedOfferings.vue
@@ -50,6 +50,10 @@ const isAnythingExpanded = computed(() => (
   || Object.values(expandedModules.value).some(Boolean)
 ));
 
+const areAllModulesCollapsed = computed(() => (
+  !Object.values(expandedModules.value).every(Boolean)
+));
+
 function areAnyComponentsExpandedIn (moduleName: string) {
   return affectsWithModuleName(moduleName).some(isExpanded);
 }
@@ -79,6 +83,12 @@ function collapseAll() {
   }
 }
 
+function expandAll() {
+  for (const moduleName in expandedModules.value) {
+    expandedModules.value[moduleName] = true;
+  }
+}
+
 function togglePsComponentExpansion(affect: ZodAffectType) {
   const isExpanded = expandedAffects.value.get(affect);
   isExpanded.value = !isExpanded.value;
@@ -102,6 +112,14 @@ function moduleComponentName(moduleName: string = '<module not set>', componentN
   <div v-if="theAffects" class="osim-affects">
     <h4 class="mb-4">
       Affected Offerings
+      <button
+        v-if="areAllModulesCollapsed"
+        type="button"
+        class="btn btn-sm btn-secondary me-2"
+        @click="expandAll()"
+      >
+        Expand All
+      </button>
       <button
         v-if="isAnythingExpanded"
         type="button"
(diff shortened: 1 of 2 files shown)


src/components/AffectedOfferings.vue
```diff
@@ -102,6 +112,14 @@ function moduleComponentName(moduleName: string = '<module not set>', componentN
   <div v-if="theAffects" class="osim-affects">
     <h4 class="mb-4">
       Affected Offerings
+      <button
+        v-if="areAllModulesCollapsed"
```
MrMarble, line 116: I feel like this conditional should be changed, e.g. if there are 15 modules and only one is expanded, it should be possible to expand the other 14 without having to manually collapse the one expanded.

Same way the https://github.com/RedHatProductSecurity/osim/pull/239/files#diff-14aca78bde51d2bb8135088852aa914ecb6c5fc2d0033143a4299ddeb20d73beL105-L112 function works, instead of checking `areAllModulesCollapsed`, check `isAnythingCollapsed`

src/components/AffectedOfferings.vue
```diff
@@ -50,6 +50,10 @@ const isAnythingExpanded = computed(() => (
   || Object.values(expandedModules.value).some(Boolean)
 ));
 
+const areAllModulesCollapsed = computed(() => (
+  !Object.values(expandedModules.value).every(Boolean)
+));
```
MrMarble, line 55:
```suggestion
const isAnythingCollapsed = computed(() => (
  !Object.values(expandedModules.value).some(Boolean)
));
```

src/components/AffectedOfferings.vue
C-Valen: Good catch, thanks for pointing out. Rechecking it I think the actual code is Ok, the problem is on the name  `areAllModulesCollapsed`, it is wrong and lead to confusions (see next comment).

src/components/AffectedOfferings.vue
C-Valen: The actual code returns true if any module is collapsed (notice the negation at the beginning of the statement). So I believe we could just correct the name to `isAnythingCollapsed` as you suggested.

I think your suggestion to change `every` to `some` would make the statement become false and hiding the button in the moment any module is expanded.

WDYT?

MrMarble: Oops, you're right, I got confused with the name and got carried away by the `isAnythingExpanded` implementation...

C-Valen: Yea the name is totally wrong and very confusing, so good job rising this!
I'll amend it soon.

MrMarble (approved):

src/components/AffectedOfferings.vue
```diff
@@ -50,6 +50,10 @@ const isAnythingExpanded = computed(() => (
   || Object.values(expandedModules.value).some(Boolean)
 ));
 
+const isAnythingCollapsed = computed(() => (
+  !Object.values(expandedModules.value).every(Boolean)
```
d10n, line 54: This is fine - I might have used Object.values(expandedModules.value).some(Boolean)

d10n (approved):

PR #249, opened 2024-06-04
Restructure features to minimize the testing data creation

RichardXuan:
Restructure features to minimize the testing data creation


features/steps/common_steps.py
```diff
@@ -33,7 +33,7 @@ def step_impl(context):
 
 @given('I go to an embargoed flaw detail page')
 def step_impl(context):
-    go_to_specific_flaw_detail_page(context.browser, True)
+    go_to_specific_flaw_detail_page(context.browser)
```
huiwangredhat, line 36: This line is the same with line 31?

features/utils.py
```diff
@@ -122,15 +123,25 @@ def go_to_home_page(browser):
...
+def go_to_specific_flaw_detail_page(browser):
     """
     Go to a specific flaw detail page
     """
-    go_to_home_page(browser)
-    home_page = HomePage(browser)
-    home_page.quickSearchBox.clear_text()
-    home_page.quickSearchBox.set_text(cve_id)
-    home_page.click_btn('quickSearchBtn')
+    cve_id = os.getenv("FLAW_ID")
```
huiwangredhat, line 130: How to get the FLAW_ID? export the FlAW_ID after finishing the flaw_create_*.feature?

features/steps/common_steps.py
RichardXuan: Yes.

features/utils.py
RichardXuan: Check https://github.com/RedHatProductSecurity/osim/blob/OSIDB-2692/features/steps/flaw_create.py#L135 and https://github.com/RedHatProductSecurity/osim/blob/OSIDB-2692/features/steps/flaw_create.py#L85 .
If you want run a single case locally, just export `FlAW_ID` you want operate.

huiwangredhat: I think if the create flaw function has something wrong, it will block all the following testing, right?

RichardXuan: Yes. It will.

huiwangredhat (approved):

PR #256, opened 2024-06-10
Feature: Internal jira comments + Comments section redesign

C-Valen:
# OSIDB-828/OSIDB-2536 Internal Jira Comments and Comments Section Redesign

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Introducing internal comments on Flaws through a Jira service to integrate external REST API.
Redesign of comments section on flaw form for an improved clarity and UX.

## Changes:

- Creates a Jira service to integrate Jira REST API:
  - Using nginx reverse proxy implementation for prod environment (props @d10n)
  - Using OSIDB forwarder endpoint for non-prod environments (props @JimFuller-RedHat)
- Creates `useInternalComments` composable and modifies `FlawComments` to support internal Jira comments
- Creates new widget `Tabs` and modifies `FlawComments` to redesign comments section on Flaw form

## Considerations:

- OsimRuntime and environment variables must be updated to include new values
- The Jira service on prod environment is using nginx reverse proxy
  For non-prod environments Jira service uses OSIDB forwarder endpoint
- I considered to make `useInternalComments` independent from `useFlawModel` just cause internal comments are not part of FlawModel and they don't have OSIDB as source.
- Timestamp on comments is displayed as UTC (System time zone)

## Demo:

https://github.com/RedHatProductSecurity/osim/assets/29104825/d2833608-a346-459d-b1f0-512d7e360977


Dockerfile
```diff
@@ -30,6 +30,8 @@ RUN microdnf --nodocs --noplugins --setopt install_weak_deps=0 -y install nginx
 # Log to symlinks in tmp
     && sed -i 's,/var/log/nginx/access.log,/tmp/logs/nginx.access.log,' /etc/nginx/nginx.conf \
     && sed -i 's,/var/log/nginx/error.log,/tmp/logs/nginx.error.log,' /etc/nginx/nginx.conf \
+# Create resolver stub file to be updated in entrypoint
+    && install -o999 -g999 /dev/null /etc/nginx/conf.d/resolvers.conf \
```
MrMarble, line 34: I would move this down along with the other `install` commands just for _aesthetics_.

https://github.com/RedHatProductSecurity/osim/blob/0a268faff816f90c33d3077a7e14440692b2768e/Dockerfile#L42-L44

build/entrypoint.d/30-reverse-proxy.sh
```diff
@@ -0,0 +1,31 @@
...
+    proxy_ssl_trusted_certificate /tmp/Proxy-CA.crt;
+    proxy_ssl_session_reuse on; # try off if connection issues
+    proxy_ssl_server_name on; # required
+EOF
+cat <<'EOF' >>/tmp/osim-nginx-proxy.conf
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto https;
+}
+EOF
```
MrMarble, line 31:
```sh
EOF
cat <<'EOF' >>/tmp/osim-nginx-proxy.conf
```
is this intended? being that we are appending to the same file I feel like we could just use only one `cat` but I'm not 100% sure what the `'EOF'` with single quotes does
```suggestion
cat <<EOF >/tmp/osim-nginx-proxy.conf
location /proxy/jira/ {
    # Trailing slash in proxy_pass strips the location directive prefix from the downstream URL
    proxy_pass ${OSIM_NGINX_PROXY_JIRA}/;

    proxy_ssl_trusted_certificate /tmp/Proxy-CA.crt;
    proxy_ssl_session_reuse on; # try off if connection issues
    proxy_ssl_server_name on; # required
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto https;
}
EOF
```

src/components/FlawComments.vue
```diff
@@ -1,149 +1,240 @@
...
 
-const activeFilters = computed(() => {
-  if (Object.values(selectedFilters.value).every((isActive) => !isActive)) {
-    return () => true;
-  }
-
-  const activeFilterFunctions = Object.entries(selectedFilters.value)
-    .filter(([, isActive]) => isActive)
-    .map(([filter]) => filterFunctions[filter as CommentFilter]);
+function parseOsidbComments(comments: any) {
```
MrMarble, line 65: We have the comment schema defined as `ZodFlawCommentSchema`, 
maybe we could add the type and use it in this component ? (not only this function, but on the props as well)
```ts
export type ZodFlawCommentSchemaType = z.infer<typeof ZodFlawCommentSchema>
```
If the schema changes, these types will be valid and will break at runtime. 

> Personally I don't like using `any` but if it's fine for the team it's fine for me

src/components/FlawComments.vue
```diff
@@ -1,149 +1,240 @@
...
+
+const newComment = ref('');
 const isAddingNewComment = ref(false);
 
 const emit = defineEmits<{
   'comment:addPublicComment': [value: any, value: any];
+  'disableForm': [value: boolean];
 }>();
 
 const SYSTEM_EMAIL = 'bugzilla@redhat.com';
```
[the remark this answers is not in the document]
superbuggy: oh wait there's actually just a `constants/` directory

src/composables/useInternalComments.ts
```diff
@@ -0,0 +1,89 @@
...
+        addToast({
+          title: 'Error',
+          body: `Error saving internal Jira comment. ${err}`,
+          css: 'warning',
+        });
+      })
+      .finally(() => isSavingInternalComment.value = false);
+  }
+
+  function internalCommentsAvailable() {
```
superbuggy, line 66: maybe use `computed` just in case the reactivity is needed at some point

superbuggy:
```js
const internalCommentsAvailable = computed(() => available.value && taskKey);
```

build/entrypoint.d/30-reverse-proxy.sh
d10n: It's intended - for reference, the `<<EOF` construct is called a heredoc (search "Here Document" in `man bash`), and shell substitutions like `$FOO` or `${FOO}` or `$(uname)` are substituted in heredocs.
If single quotes are used around the heredoc label, like `<<'EOF'`, then the heredoc is literal and no substitutions are made.
To substitute the environment variable `${OSIM_NGINX_PROXY_JIRA}`, the first heredoc is unquoted, and to use the nginx config variables like `$remote_addr`, the second heredoc is quoted. I just had a weak preference to split the heredocs instead of escaping the nginx config variables like `\$remote_addr`

src/components/FlawForm.vue
```diff
@@ -487,10 +488,11 @@ const expandFocusedComponent = (parent_uuid: string) => {
       <div v-if="mode === 'edit'" class="border-top osim-flaw-form-section">
         <FlawComments
           :comments="flaw.comments"
+          :taskKey="flaw.task_key"
           :error="errors.comments"
           :isSaving="isSaving"
           @comment:addPublicComment="addPublicComment"
-          @refresh:flaw="emit('refresh:flaw')"
+          @disableForm="(value) => formDisabled = value"
```
superbuggy, line 495: I think here you could just use`v-model:isSaving` and avoid having to introduce another piece of state and emitting

Dockerfile
d10n: I put it here because it's grouped with the other /etc/nginx lines, but I don't have a strong opinion on where it goes

superbuggy (approved):

src/components/FlawForm.vue
superbuggy: I tested out the change and it's not significant, and I don't feel strongly

tboyetteredhat (approved):

build/entrypoint.d/30-reverse-proxy.sh
MrMarble: I knew about the `<<EOF` syntax for writing files but the `<<'EOF'` is an interesting find, thanks for the explanation!

Dockerfile
MrMarble: It's just because it starts with `install -o999 -g999 /dev/null`, doesn't really matter 😆

src/components/FlawComments.vue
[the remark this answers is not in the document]
C-Valen: I don't see it proper to move it into `useInternalComments` cause it is used on system comments.

Having it on `constants/` doesn't convince me much neither as the value is used only in this template currently. If it'd be required in other components, I'll say to create a `global.ts` file on `constants/`, but doesn't makes sense to me having that file with only this value considering it is not globally required. 

I think we can leave it there for now, as it in the same scope as it's used.

PR #258, opened 2024-06-10
Add new affected component button

tboyetteredhat:
# [[OSIDB-2876](https://issues.redhat.com/browse/OSIDB-2876)] [Add new affected component button]
# [[OSIDB-2533](https://issues.redhat.com/browse/OSIDB-2533)] [Sort affects by product family, alphabetical]

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

OSIDB-2876:  Add new affected component button
OSIDB-2533: Sort affects by product family, alphabetical

## Considerations:

Closes OSIDB-2876
Closes OSIDB-2533

[Screencast from 2024-07-05 07-35-59.webm](https://github.com/RedHatProductSecurity/osim/assets/156366006/70bfd89c-d552-44e7-8ab9-07e39267b9a9)


C-Valen (commented):
Some suggestions for possible UI/UX improvements:
- Make buttons a bit smaller so they are consistent with other buttons in the section (e.g. same size as expand/collapse, file tracker etc.)
- Auto expand added modules and components
- When adding a component, all other modules remain in same collapsed/expanded state. (currently all other expanded modules are collapsed when a new module is added)

tboyetteredhat:
> Some suggestions for possible UI/UX improvements:
> 
> * Make buttons a bit smaller so they are consistent with other buttons in the section (e.g. same size as expand/collapse, file tracker etc.)
> * Auto expand added modules and components
> * When adding a component, all other modules remain in same collapsed/expanded state. (currently all other expanded modules are collapsed when a new module is added)

Thanks for your feedback. I will fix those issues.

C-Valen (commented):
I think we could avoid using modules names for indexing. 

It also would be nice to validate names so the users get a feedback for example when they are using a name that already exists (even in draft module) before saving or when the module name is empty.

Minimum relevance but the position for the module name field feels a bit of unnatural to me, I would have preference to have it inside the collapse (if there are not drawbacks). I wonder if would it be easy to just make the module name editable when clicked, and on blur saving it and make it static, avoiding that way the use of additional fields and space.

src/components/AffectedOfferings.vue
```diff
@@ -17,11 +18,14 @@ const props = defineProps<{
...
 
+const uniqueModuleId = ref(0);
+
 const emit = defineEmits<{
   'file-tracker': [value: object];
   'affect:remove': [value: ZodAffectType];
   'affect:recover': [value: ZodAffectType];
-  'add-blank-affect': [];
+  'add-blank-affect': [value: string, callback: (affect: ZodAffectType) => void];
+  'update-module-name': [previousModuleName: string, newModuleName: string];
```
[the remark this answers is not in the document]
tboyetteredhat: I think if we update moduleName, we need to update all affects under that moduleName. Those affects are un saved in OSIDB and just new affects

src/composables/useFlawAffectsModel.ts
```diff
@@ -96,18 +96,17 @@ export function useFlawAffectsModel(flaw: Ref<ZodFlawType>) {
     affectsToDelete.value = [];
   }
 
-  function addBlankAffect() {
+  function addBlankAffect(moduleName: string, callback: (affect: ZodAffectType) => void) {
```
superbuggy, line 99:
```suggestion
  function addBlankAffect(): ZodAffectType {
    const embargoed = flaw.value.embargoed;
    flaw.value.affects.push({
      embargoed,
      affectedness: '',
      resolution: '',
      delegated_resolution: '', // should this be null
      ps_module: '',
      ps_component: '',
      impact: '',
      cvss_scores: [{
        // affect: z.string().uuid(),
        cvss_version: 'V3',
        issuer: 'RH',
        comment: '',
        score: null,
        vector: '',
        embargoed,
        alerts: [],
      }],
      trackers: [{ errata: [] }],
      alerts: [],
    } as ZodAffectType);
    return flaw.value.affects.at(-1);
  }
```

superbuggy: This would avoid having to add a callback by just returning the new affect that was just added to the flaw's list of affects, this should be reactive

src/components/AffectedOfferings.vue
```diff
@@ -118,6 +126,36 @@ function moduleComponentName(moduleName: string = '<module not set>', componentN
...
+  return expandedState && (!initialAffectedModules.value.includes(moduleName)
+  || affectsWithModuleName(moduleName).every(affect => !affect.uuid));
+}
+
+function addNewModule() {
+  collapseAll();
+  uniqueModuleId.value += 1;
+  const newModuleName = `New Module ${uniqueModuleId.value}`;
+  affectedModules.value.push(newModuleName);
+  expandedModules.value[newModuleName] = true;
```
superbuggy, line 140: here you could just call the modified `addBlankAffect` function (outlined below, the change being to return the new blank affect) and modify the `.ps_module` property of the affect.

superbuggy (changes requested):

MrMarble:
Could you add some screenshots or a video, maybe to the jira ticket, to understand the changes better?

superbuggy:
@tboyetteredhat and I had a call and it seemed like the best course of action might be to split these issues up and to handle sorting separately from changing the affects UI which @C-Valen is redesigning--there may be a way to avoid the user having to deal with the screen jumping around when the affect's module and component are set (a consequence of the sorting of affects)

PR #271, opened 2024-06-12
Fix OSIDB-2904: Unembargoed btn not shown

C-Valen:
# OSIDB-2904 Unembargo button not shown on embargoed flaws

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [-] No changelog updated
- [x] Jira ticket updated

## Summary:

Fix the bug where unembargo button was not appearing on flaw form of embargoed flaws.
The problem was that removing the ref on `initialFlaw` made the `isEmbargo` not being reactive, so the embargoed component was reviving undefined for that `isEmbargoed` property.

## Changes:

- Make isEmbargoed a ref and edit the values every time `initialFlaw` changes
- Remove non required `reactive` on `initialFlaw` as it is nor reactive anymore

## Considerations:

- isEmbargoed need to be reactive as it is required to make the embargoed field changing after the flaw refresh

### DIFF
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -81,6 +81,7 @@ let initialFlaw: ZodFlawType;
 
 onMounted(() => {
   initialFlaw = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  isEmbargoed.value = initialFlaw?.embargoed;
   if (draftFlaw) {
     flaw.value = useDraftFlawStore().addDraftFields(flaw.value);
     useDraftFlawStore().$reset();
@@ -89,10 +90,11 @@ onMounted(() => {
 
 watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any change (though that shouldn't happen)
   initialFlaw = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  isEmbargoed.value = initialFlaw?.embargoed;
   onReset();
 });
 
-const isEmbargoed = computed(() => initialFlaw?.embargoed);
+const isEmbargoed = ref();
 const showUnembargoingModal = ref(false);
 const unembargoing = computed(() => isEmbargoed.value && !flaw.value.embargoed);
 
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { DateTime } from 'luxon';
-import { computed, ref, watch, onMounted, reactive } from 'vue';
+import { computed, ref, watch, onMounted } from 'vue';
 import { deepCopyFromRaw } from '@/utils/helpers';
 
 import LabelEditable from '@/components/widgets/LabelEditable.vue';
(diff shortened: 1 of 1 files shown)


tboyetteredhat (approved):

src/components/FlawForm.vue
```diff
@@ -89,10 +90,11 @@ onMounted(() => {
 
 watch(() => props.flaw, () => { // Shallow watch so as to avoid reseting on any change (though that shouldn't happen)
   initialFlaw = deepCopyFromRaw(props.flaw) as ZodFlawType;
+  isEmbargoed.value = initialFlaw?.embargoed;
   onReset();
 });
 
-const isEmbargoed = computed(() => initialFlaw?.embargoed);
+const isEmbargoed = ref();
```
MrMarble, line 97: I would add the default `false` value just to prevent any `falsy` check, but is not strictly needed here
```suggestion
const isEmbargoed = ref(false);
```

superbuggy:
```
const isEmbargoed = () => initialFlaw.embargoed;
```

Is your approach preferable to the one above? Suggesting for the sake of avoiding proliferating reactive state where possible.

MrMarble: @superbuggy The base problem is that `initialFlaw` is a normal variable, not a reactive one, so any change to the variable is not propagated, that's why `isEmbargoed` was set to `undefined` and not being updated after the `onMounted` hook.

Using a function instead of a reference has the same problem, it might work now but any change to the event loop of vue can break it, see this [playground](https://play.vuejs.org/#eNqNVNtu2zAM/RVOL0vRINm6PQWJd0MLbMAu6No3oYBq045aWTJ0SVoE/vdRkuMmRVHsxZDJQ+rwkNSOfem62SYgW7ClK63sPDj0oSu4lm1nrIcdWKynsBW+XE/B6J8maI8V9FBb08JbCn7LNdcKPdRKbOO5NNp5uBUOYRXDJ+9O9kbpzttbYRtDKVYwOYFVkcI+zXBvfwF6ifWQqRbKIWXjeqQyyVl2XENKRch0BghBVovEY7YRKuA0m8ebFuBtwGjs42c+P74yB1G6mPWQX4TCNdFsW9Re6gb8Wjq4ubkZfFupFIiqAgGcdVE6zqB8LBWCqD1awiO0kf4UGrnJGfCwYqqpDrr00mjwskXwBkJXCZ/o0hXaeFkirM2WwiDoCmupSdTb4KlL5IkJB4YRYbE1G/LTcQ+l1IkDTB5NgFJoAikjqhTaiSbq3CetU/cHnQ/kHPo3Kj+LgpNeT5Do+U9R801cL+d5EGkE6cdj2ymqmv4Aluv3xQWFLWC3y63u++WcjNl5VlyQZMl5rGRCnY0oYvEcFAfsEEUqepLnc6lkeb/iLFZ0espZcZ16QMJG3ShHGvIYmSNejj4ulWp/83zkWeFN0yjE40RdcUWtyC5ql3qMO6Cw9C63dy10g26YKUFOQROzwYhCi3EK4mwkyhUZKyu2+1mj0eiMpuFYzuO2L+cHSrMp846WsJbN7M4ZTa9D6jFnMUoqtL+7OJmOMxIyLxVnQimz/ZFsca2GZaOYNZb3L9jv3EO0cfbHokO7Qc5Gnydd0Gf3+d9f+EDn0dmaKihCv+K8RGdUiBwz7CstCNE+wCW239MbR5JcufMHj9rtixrfhYTnjF65b6+U/kT3w+zj8J70rP8H5NLU1g==) as an example.

I'm not very experienced with vue so if I'm wrong please correct me 😆

C-Valen: @superbuggy after some tests I noticed the function approach breaks the reactivity for updating the embargoed field after finishing an unembargo workflow. This is hiding the button just after a flaw has been unembargoed.

PR #285, opened 2024-06-14
Task: Unified comments model

C-Valen:
#  Unified comments model

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Unifies all comments logic into a single FlawCommentsModel composable, drastically reducing complexity and improving code maintainability.

## Changes:

- Minor function/type/parameter name corrections
- Separated comments model logic from component template
- Relocate handlers to comment services
- Refactor `useInternalComments` and `addFlawComment` (from `flawModel`) into `useFlawCommentsModel`
- Moves info message in comments tabs above the new comment input

## Considerations:

Depends on https://github.com/RedHatProductSecurity/osim/pull/284


src/components/FlawForm.vue
```diff
@@ -506,12 +513,17 @@ const formDisabled = ref(false);
...
-          @disableForm="(value) => formDisabled = value"
+          :publicComments
+          :privateComments
+          :internalComments
+          :internalCommentsAvailable
+          :isLoadingInternalComments
+          :systemComments
+          :taskKey="flaw.task_key || ''"
+          :bugzillaLink
+          :isSaving
```
MrMarble, line 524: What do you guys think about using a `v-bind` here to reduce the number of props? ([docs](https://vuejs.org/api/built-in-directives.html#v-bind))
Instead of returning all `xxxxxComments` props from the `useFlawModel`, return an object like `flawComments` that wraps all and pass that to the component.
With `v-bind` the object automatically expand to the component props
```js
// Inside useFlawModel
const flawComments = {
  publicComments,
  privateComments,
  internalComments,
  systemComments,
  internalCommentsAvailable,
  isLoadingInternalComments,
}

return { 
flawComments, 
addFlawComment,  
loadInternalComments,
// rest...
}
```
```js
// inside FlawForm
const {
flawComments,
addFlawComment,
loadInternalComments
// rest...
} = useFlawModel(props.flaw, onSaveSuccess);
```

```vue
<FlawComments v-bind="flawComments"
  :taskKey="flaw.task_key || ''"
  :bugzillaLink
  :isSaving
  @comment:addFlawComment="addFlawComment"
  @loadInternalComments="loadInternalComments"
/>
```

Function wise is the same but maybe a bit cleaner? I would like to know your opinion on this

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * Create and delete Affects in single request (`OSIDB-2821`)
 * Add link to bugzilla tracker on Flaw form (`OSIDB-2897`)
 * Set public date to current date on unembargo (`OSIDB-2829`)
+* Support private Bugzilla comments (`OSIDB-2912`)
+* Add button to Bugzilla on public and private comments
```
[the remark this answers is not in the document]
C-Valen: There is no task related to this

src/components/FlawForm.vue
C-Valen: That's a good point. I think that approach is adding a bit of complexity, I'd weakly prefer to have some visibility on different props a component is receiving over wrapping them too much.

What I thought to do in order to reduce the props in this case is to use a single `flawComments` array on `useFlawCommentsModel` which contains all the comment types, and compute the individual comments arrays by type within the `FlawComments` template.

The reason I didn't do that is just in case want to add some new logic to a specific comment type only, in which case I think it would be proper to have it in the model.

src/components/FlawComments.vue
```diff
@@ -66,79 +57,62 @@ const handleTabChange = (index: number) => {
...
+
+function linkify(text: string) {
   const bugzillaLink = `${osimRuntime.value.backends.bugzilla}/show_bug.cgi?id=`;
 
-  const urlRegex = /\b(https?:\/\/[\S]+)\b/g;
-  const jiraTagRegex = /\[~([^[\]]+)\]/g;
   const bugzillaRegex = /\[bug (\d+)\]/g;
-  return (
-    sanitizeHtml(text)
-      .replace(urlRegex, '<a target="_blank" href="$1">$1</a>')
```
d10n, line 138: Can we add this back in, or is there a reason to remove it?

src/services/JiraService.ts
```diff
@@ -140,7 +141,10 @@ export async function postJiraComment(taskId: string, comment: string) {
     data: {
       body: comment
     },
-  }).then((response) => response.data);
+  })
+    .then(createSuccessHandler({ title: 'Success!', body: 'Internal comment saved' }))
+    .then((response) => response.data)
+    .catch(createCatchHandler('Error saving internal Jira comment'));
```
MrMarble, line 147: Why are we handling errors here? This is the only service that handles error internally, making the use a bit confusing.

`createSuccessHandler` and `createCatchHandler` uses the `toast` store to show the message to the screen, I think services should be side-effects free, so they can be used anywhere, is up to the caller to handle the errors, so they can decide what to do (fail silently, retry on error...).

src/services/FlawService.ts
```diff
@@ -136,18 +136,21 @@ export async function postFlawCvssScores(flawId: string, cvssScoreObject: unknow
...
     data: {
       text: comment,
       creator: creator,
       is_private: isPrivate,
       embargoed,
     },
-  }).then((response) => response.data);
+  }).then(createSuccessHandler({ title: 'Success!', body: `${type} comment saved.` }))
+    .then((response) => response.data)
+    .catch(createCatchHandler(`Error saving ${type} comment`));
```
MrMarble, line 153: I have the same opinion here, but I see we do handle errors on all services of this file, so maybe we should decide if we handle errors inside each service or not, to keep consistency between services

src/components/FlawForm.vue
```diff
@@ -511,12 +518,17 @@ const createdDate = computed(() => {
...
           @comment:addFlawComment="addFlawComment"
-          @disableForm="(value) => formDisabled = value"
+          @loadInternalComments="loadInternalComments"
```
MrMarble, line 531: And this then makes sense to use the `v-bind` directive to reduce the footprint, we can also use the `v-on` directive to bind multiple events if needed
```html
<FlawComments v-bind="commentsModel" v-on="commentsEvents"  :taskKey="flaw.task_key || ''" />
```

src/composables/useFlawCommentsModel.ts
```diff
@@ -0,0 +1,107 @@
...
+  const systemComments = computed<ZodFlawCommentType[]>(() => flaw.value.comments.filter(filterFunctions.system));
+
+  type OsidbCommentFilter = 'public' | 'private' | 'system';
+  type OsidbCommentFilterFunctions = Record<OsidbCommentFilter, (comment: ZodFlawCommentType) => boolean>;
+
+  const filterFunctions: OsidbCommentFilterFunctions = {
+    public: (comment: any) => !comment.is_private,
+    private: (comment: any) => comment.is_private && comment.creator !== SYSTEM_EMAIL,
+    system: (comment: any) => comment.creator === SYSTEM_EMAIL,
+  };
```
MrMarble, line 24: This will be recreated each time the `useFlawCommentsModel` function is called, since they don't use any variable from the `useFlawCommentsModel` I think we can move them outside of the body of the function

src/composables/useFlawCommentsModel.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+        internalCommentsAvailable.value = false;
+      })
+      .finally(() => {
+        isLoadingInternalComments.value = false;
+      });
+  }
+
+  function addFlawComment(comment: string, creator: string, type: string) {
+    if (type !== 'Internal') {
```
MrMarble, line 58: We have defined the comment types `Intenal, Public, Private, System` multiple times each time with different casing (`Public` vs `public`), I think it makes sense to have them in a constant and use it everywhere (like the enum in `FlawForm`

```typescript
const enum COMMENT_TYPES {
  PUBLIC = "Public",
  INTERNAL= "Internal",
  PRIVATE = "Private", 
  SYSTEM = "System"
}
```
This also provides a way to automatically calculate the type on https://github.com/RedHatProductSecurity/osim/blob/b1336d29423d1301f34abea96785053f1e64d216/src/composables/useFlawCommentsModel.ts#L17

```typescript
type OsimCommentFilter = Lowercase<Exclude<keyof typeof COMMENT_TYPES, "INTERNAL">>
//		^ = "public" | "private" | "system"
```

<details>
<summary>Always prefer constant enums</summary>

On typescript, when you use an enum
```typescript
enum Sample {
  VALUE1
  VALUE2
}

console.log(Sample.VALUE1);
```
it will create a runtime function for checking the values:

```js
var Sample;
(function (Sample) {
    Sample[Sample["VALUE1"] = 0] = "VALUE1";
    Sample[Sample["VALUE2"] = 1] = "VALUE2";
})(Sample || (Sample = {}));

console.log(Sample.VALUE1);
```

but if the enum is declared as constant (`const enum Sample { ... }`) it will just replace the usage with the constant value on build:
```js
console.log(0 /* Sample.VALUE1 */);
```
See the example in [Typescript Playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBApmArgWxgYQPIFksFEByAKgPqECaACrgMrGb7WEwDeAsAFAwwUCqAQgBkAkmhgBeGACIKiAEYAbAJbBJAGg5chRXACV8AQQETJQsFDgAnMAEN5ajdx1CAavsK5xUihcUA3a+ZqMA7UZIy4WJ6S1ACe0HDIkhwAvhwcCCjo2HhEpJQ0LA68giJRMgrK9pwwWu56hsam5la2VVwUTq7uZT7+garB1aHhkcax8YkpaewA9DMwoJBQ1mYw-vKIcDDgMLKIivIAJhyLECDycAB08iAA5gAUmDgEJORUtPSMl7W6BgIAlNM5jALIgzIpkFsAGZg4BQRTgE7gM4Xa53R7ZF55d7fbT1AEcIA)
</details>

superbuggy (approved):

MrMarble (approved):

PR #287, opened 2024-06-17
Add test case for filing tracker to jira with affectness is Affected

huiwangredhat:
# [OSIDB-1962] [Create Trackers on Jira For affected with Affectedness=AFFECTED]


RichardXuan (approved):
LGTM

PR #288, opened 2024-06-17
Fix OSIDB-2827: Flaw list text overlap

C-Valen:
# OSIDB-2827 Flaw list text overlap

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Fixes text overlaps on flaw queue when status value is `PRE_SECONDARY_ASSESSMENT` by making the table larger in width.
Additionally adds a couple of minor UI/UX enhancements that were considered after this change.

## Changes:

- Increases flaw queue table's width
- Adjust spacing on navbar to for aesthetics
- Make loading/loaded flaws indicator act like a single element
- Removes unused code

### Width adjustments:
- Before:
OSIM-FlawQueue-Before

- After:
OSIM-FlawQueue-After

### Loading/Loaded:
- Before:
[OSIM-LoadingIndicator-Before.webm](https://github.com/RedHatProductSecurity/osim/assets/29104825/bff773cf-8e64-4893-accb-59089775a235)

- After:
[OSIM-LoadingIndicator-After.webm](https://github.com/RedHatProductSecurity/osim/assets/29104825/3ffc88fb-81b8-40aa-9203-0ff421b5e626)

## Considerations:

- Loading/loaded flaws indicator is covered if there are some alerts, may be we may consider relocating
- There is a to-do task to remove whitespaces on the application (OSIDB-2005) the idea is to adjust the width of other views (like flaw form or advanced seatch) on that task so all the UI has a consistent layout within different views
- I think it would be nice to have embargoed badge on an individual table column, for better visibility and equal height of all rows. Also probably applying a truncate on long titles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 * Flaws without a Jira task cannot be updated (`OSIDB-2960`)
 * Remove Trackers section on Flaw Edit (`OSIDB-2954`)
+* Owner - Status text overlap on flaw list (`OSIDB-2827`)
 
 ## [2024.6.1]
 ### Added
--- a/src/components/IssueQueue.vue
+++ b/src/components/IssueQueue.vue
@@ -178,7 +178,15 @@ const nameForOption = (fieldName: string) => {
         <span> Loading Flaws&hellip; </span>
       </div>
       <span
-        v-else-if="issues.length"
+        v-if="isLoading"
+        class="spinner-border spinner-border-sm d-inline-block ms-3"
+        role="status"
+      >
+        <span class="visually-hidden">Loading...</span>
+      </span>
+      <span v-if="isLoading"> Loading Flaws&hellip; </span>
+      <span
+        v-if="issues.length"
         class="float-end"
         :class="{'text-secondary': isLoading}"
       > Loaded {{ issues.length }} of {{ total }}</span>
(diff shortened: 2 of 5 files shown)


MrMarble:
Maybe we could remove the margins everywhere like OSIDB-2005 ask for instead of fixing this and later in time fixing the rest of the views?

This will introduce a "jump" in margins when navigating to/from the flaw form

C-Valen:
> Maybe we could remove the margins everywhere like OSIDB-2005 ask for instead of fixing this and later in time fixing the rest of the views?
> 
> This will introduce a "jump" in margins when navigating to/from the flaw form

OSIDB-2005 Is an extension of this (It depends on this PR). I separated cause are two different scopes, this is the specific fix for the OSIDB-2827 issue so I considered it worth to have it in an independent PR.

Also OSIDB-2005 applies many other style improvements so the fix could be difficult to be tracked there.

No strong preference, if we all agree we can skip this and go directly with OSIDB-2005 as it already contains this fix.

MrMarble:
> OSIDB-2005 Is an extension of this (It depends on this PR). I separated cause are two different scopes, this is the specific fix for the OSIDB-2827issue so I considered it worth to have it in an independent PR.
> 
> Also OSIDB-2005 applies many other style improvements so the fix could be difficult to be tracked there.
> 
> No strong preference, if we all agree we can skip this and go directly with OSIDB-2005 as it already contains this fix.

Yes I don't mind if we merge this PR until we have time to work on OSIDB-2005, just asking if it makes sense to do the 2005 directly

JakubFrejlach (approved):
LGTM!

tboyetteredhat (approved):

superbuggy (approved):

PR #289, opened 2024-06-17
OSID-2933: Fix outdated model instance error

MrMarble:
# OSID-2933: Fix outdated model instance error

## Checklist:

- [ ] Linting passed
- [ ] Type checks passed
- [ ] Tests suite passed
- [ ] Commits consolidated
- [ ] Changelog updated
- [ ] Jira ticket updated

## Changes:

## Considerations:


superbuggy (approved):

CHANGELOG.md
```diff
@@ -58,6 +58,7 @@
 * Unable to save new references and ackowledgments on flaws (`OSIDB-2206`)
 * Reset affects when flaw is reset (`OSIDB-2793`)
 * New public comments are not displayed (`OSIDB-2700`)
+* `updated_dt` field should be updated on flaw save (`OSIDB-2933`)
```
C-Valen, line 61: We aim to have changelog as user friendly as possible.
For example for this case I'd add something like:
```suggestion
* Fix 409 errors on saving affects  (`OSIDB-2933`)
```

d10n (approved):

C-Valen (approved):
🚀

CHANGELOG.md
superbuggy: Ah yeah, good point, or even 'Fix outdated model (409) errors on saving affects'

MrMarble: > We aim to have changelog as user friendly as possible.
> For example for this case I'd add something like:
> ```suggestion
> * Fix 409 errors on saving affects  (`OSIDB-2933`)
> ```

Yeah that's a bit more user-friendly 😂
Will change the message and merge tomorrow morning

MrMarble:
> Thank you for the quick fix! @MrMarble I checked out the codes and did a test locally. I didn't encounter the 409 outdated model error, I could add new affect as expected. However, I noticed I failed to update the new added affect without a page refresh(affect update works well after page refresh), the bug could be marked as verified if this is acceptable. WDYT?

I did some testing and noticed that after a save, you can't edit any affect, not only the one you added, that functionality seems to be broken and as far as I can see is not related to the `409` error. 
I like to keep fixes to a minimum so they can be tracked easily in case a revert/cherrypick is needed, I'll create a new jira task for it, thanks!

PR #293, opened 2024-06-18
Use correct Jira username for Flaw owner

JakubFrejlach:
# [OSIDB-2715 OSIDB-2716] [Correct Jira username]

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

This PR fixes Flaw Owner Self assign and filtering for My Issues on index page. Jira username is fetched on Login or when Jira token is correctly set.

## Changes:

* fetching Jira username from Jira as usernames in Jira have various formats (emails, kerberos names, etc.)
* storing Jira username in userStore
* using the correct Jira user name for Flaw Owner self assign and My Issues filter

## Considerations:
n/a

### DIFF
--- a/src/components/FlawFormOwner.vue
+++ b/src/components/FlawFormOwner.vue
@@ -1,20 +1,32 @@
 <script setup lang="ts">
 import EditableText from '@/components/widgets/EditableText.vue';
 import { useUserStore } from '@/stores/UserStore';
-import { nextTick, computed } from 'vue';
+import { nextTick, computed, ref } from 'vue';
+
 const owner = defineModel<string | null>({ default: null });
 const userStore = useUserStore();
 
-function selfAssign() {
-  owner.value = userStore.userName;
+async function selfAssign() {
+  isLoading.value = true;
+  return userStore.updateJiraUsername().then(() => {
+    if (userStore.jiraUsername !== '') {
+      owner.value = userStore.jiraUsername;
+    }
+  }).finally(() => {
+    isLoading.value = false;
+  });
 }
 
-function handleClick(fn: (arg?: any) => any) {
-  selfAssign();
+async function handleClick(fn: (arg?: any) => any){
+  await selfAssign();
   nextTick(fn);
 }
 
-const isAssignedToMe = computed(() => owner.value === userStore.userName);
+const isAssignedToMe = computed(() =>
+  owner.value === userStore.jiraUsername && userStore.jiraUsername !== ''
+);
+
+const isLoading = ref(false);
 
 </script>
 
@@ -26,12 +38,22 @@ const isAssignedToMe = computed(() => owner.value === userStore.userName);
       </span>
       <EditableText v-model="owner">
         <template v-if="!isAssignedToMe" #buttons-out-of-editing-mode="{ onBlur }">
-          <button type="button" class="btn btn-primary osim-self-assign" @click.prevent.stop="handleClick(onBlur)">
+          <button
+            type="button"
+            class="btn btn-primary osim-self-assign"
+            :disabled="isLoading"
+            @click.prevent.stop="handleClick(onBlur)"
+          >
             Self Assign
           </button>
         </template>
         <template v-if="!isAssignedToMe" #buttons-in-editing-mode="{ onBlur }">
-          <button type="button" class="btn btn-primary osim-self-assign" @click.prevent.stop="handleClick(onBlur)">
+          <button
+            type="button"
+            class="btn btn-primary osim-self-assign"
+            :disabled="isLoading"
+            @click.prevent.stop="handleClick(onBlur)"
+          >
             Self Assign
           </button>
         </template>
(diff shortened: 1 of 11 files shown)


d10n (changes requested):
Requesting some changes, will review tomorrow

src/components/FlawFormOwner.vue
```diff
@@ -1,20 +1,28 @@
...
+
 const owner = defineModel<string | null>({ default: null });
 const userStore = useUserStore();
 
-function selfAssign() {
-  owner.value = userStore.userName;
+async function selfAssign() {
+  isLoading.value = true;
+  await userStore.cacheJiraUsername();
+  owner.value = userStore.jiraUsername;
```
MrMarble, line 12: if `userStore.cacheJiraUsername()` fails, because it has a `try/catch` it will not throw and the `owner.value` will be set to `''`, removing the previous owner if there was any

src/components/IssueQueue.vue
```diff
@@ -52,7 +56,16 @@ const params = computed(() => {
   const paramsObj: Record<string, any> = {};
 
   if (isMyIssuesSelected.value) {
-    paramsObj.owner = userStore.userName;
+    if (userStore.jiraUsername === '') {
+      addToast({
```
MrMarble, line 60: This message is already part of the `cacheJiraUsername` function that is called after login, this could lead to a double notification

MrMarble:
The function `cacheJiraUername` is called on login and caches the result on localStorage.
I think we could remove all the subsequent calls to that function and only access the `userStore.jiraUername`, as the jira username will be set on login and saved forever (until logout, which will save it again on the next login) or fail until the user fixes/sets the jira api key.

src/stores/UserStore.ts
```diff
@@ -89,6 +94,28 @@ export const useUserStore = defineStore('UserStore', () => {
...
+        addToast({
+          'title': 'Jira Error',
+          'body': `Error getting your username from <a href="${osimRuntime.value.backends.jiraDisplay}">Jira</a>. ` +
+          'Is your Jira Access Token valid?',
+          'bodyHtml': true,
+          'css': 'warning',
+        });
+      }
+    }
+  }
```
C-Valen, line 118: Is not the `cacheJiraUsername` expendable? For example having the logic of checking/getting the Jira username on `jiraUsername`?

Regarding the error handling I think is preferable to do it in the service, cause it separates the concerns between the store stuff and the fetching logic.
If that's the case may be it also have sense to make the error message more generic, so if other component uses a Jira username fetch from the service the message is also valid even if the context in the application is different.

src/stores/UserStore.ts
```diff
@@ -40,15 +42,18 @@ const userStoreLocalStorage = z.object({
   refresh: z.string(),
   env: z.string(),
   whoami: whoamiResponse.nullable(),
+  jiraUsername: z.string(),
```
C-Valen, line 45: I think this can be good scenario to create a user model with other fields we get from Jira.

MrMarble: I added a `zodJira.ts` file on my PR https://github.com/RedHatProductSecurity/osim/pull/306/files#diff-4649669d92d9bc235fd085e35b7e1f000b295a3c5c1737007ab5e0a62b383d0a with the responses of the endpoints I use, I think it makes sense to type the `getJiraUserInfo` as well

src/stores/UserStore.ts
JakubFrejlach: Is it possible to have a async computed, I read somewhere that async computed is only working with some plugin. I was actually initially thinking about caching the jiraUsername when accessing it in the storage, but I wasn't sure whether it should do such side effects, wdyt?

Agreed about the error message, will move it to the Jira service.

JakubFrejlach:
> The function `cacheJiraUername` is called on login and caches the result on localStorage. I think we could remove all the subsequent calls to that function and only access the `userStore.jiraUername`, as the jira username will be set on login and saved forever (until logout, which will save it again on the next login) or fail until the user fixes/sets the jira api key.

So in such case, we need to throw some error message when user is using Self Assign or My issues button and the Jira username is not cached, right?

d10n (approved):
Approved pending updates

src/components/FlawFormOwner.vue
```diff
@@ -1,20 +1,32 @@
...
+import { nextTick, computed, ref } from 'vue';
+
 const owner = defineModel<string | null>({ default: null });
 const userStore = useUserStore();
 
-function selfAssign() {
-  owner.value = userStore.userName;
+async function selfAssign() {
+  isLoading.value = true;
+  userStore.updateJiraUsername().then(() => {
```
d10n, line 11: Return or await the promise chain to preserve the async behavior
```suggestion
  return userStore.updateJiraUsername().then(() => {
```

PR #305, opened 2024-06-21
Add empty workflow state in Flaws

dmonzonis:
# [OSIDB-2960] [Flaws without a Jira task cannot be updated]

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

In OSIDB's commit 59884a2839465db2b5e2ed3356a70b2dea2fbdd0 the empty workflow state was added. This is not taken into account in OSIM, so when trying to update a flaw that has an empty string as workflow state, the type check will not even allow the call to OSIDB to be made.

This commit fixes this by adding the empty string as a possible workflow state.

Closes OSIDB-2960.

## Changes:

## Considerations:


MrMarble:
Just for reference, I had this fix ready as part of my PR #297 for creating jira tasks on legacy flaws which is blocked right now.
I'll remove the code from my PR and we can merge this small fix instead

src/generated-client/api.ts
MrMarble, line 1: I would generate the client with `yarn generate-openapi-client` instead of manually changing the file, as this being auto generated, could be overridden without notice.
Otherwise it LGTM

MrMarble (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 ### Added
 * Support private Bugzilla comments (`OSIDB-2912`)
 
+### Fixed
+* Flaws without a Jira task cannot be updated (`OSIDB-2960`)
```
C-Valen, line 8: Can we move this to a new `[Unreleased]` block

C-Valen (changes requested):

src/generated-client/api.ts
dmonzonis: I did not know about this command. However I tried running it locally and it modified a lot of stuff, as it looks like the generated client I had was from version 3.7.3 and the OpenAPI schema is at 4.0.0 already. I doubt this is desirable in this case, is there an easy way to limit it to my field only?

src/generated-client/api.ts
```diff
@@ -2409,6 +2409,7 @@ export interface FlawClassification {
 }
 
 export const FlawClassificationStateEnum = {
+    NoValue: '',
```
d10n, line 2412: This is a generated file, so when it is regenerated, this change will be lost.
Is there a place in `src/types/` to update zod definitions that will work instead?

d10n (changes requested):
Discussion needs to be resolved

src/generated-client/api.ts
MrMarble: I think it makes sense as the OSIDB change was made for the 4.0.0 release https://github.com/RedHatProductSecurity/osidb/releases/tag/4.0.0 but if you think this goes outside the scope of the fix, we can leave the update for later

dmonzonis: I am okay with updating it all to 4.0.0, but I will leave it in a different commit for granularity

src/generated-client/api.ts
dmonzonis: I generated the client in a different commit so this should not be an issue anymore

d10n (approved):

tboyetteredhat (approved):

PR #306, opened 2024-06-21
OSIDB-2916: Add Contributors field

MrMarble:
# OSIDB-2916: Add Contributors field

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Added the `Contributors` field from Jira to OSIM

image
> Censored for privacy reasons. There's an uncensored video attached to the jira ticket

## Changes:

~While working on this task I got the `[object Object]` error and added the missing `return` that fixes the issue when osidb returns html (i.e stack trace of error).~ Fixed in https://github.com/RedHatProductSecurity/osim/pull/308

Created new `FlawContributors` component, is similar to the `TagsInput` widget, decided not to reuse the component because the auto-suggestions behavior is hard to implement in a reusable way. All the logic is inside the `useJiraContributors` composable.

I've added a generic type parameter to the `jiraFetch` function in `JiraService` so we can type the response easily, as well as some basic zod schemas and types for the response of the 2 new methods I added

## Considerations

https://github.com/RedHatProductSecurity/osidb/pull/598 **needs** to be merged/deployed first for editing the jira task to work

### DIFF
--- a/src/components/FlawContributors.vue
+++ b/src/components/FlawContributors.vue
@@ -1,0 +1,163 @@
+<script setup lang="ts">
+import { onBeforeMount, ref } from 'vue';
+import { isDefined, watchDebounced } from '@vueuse/core';
+import sanitize from 'sanitize-html';
+import LabelDiv from './widgets/LabelDiv.vue';
+import useJiraContributors from '@/composables/useJiraContributors';
+import type { ZodJiraUserPickerType } from '@/types/zodJira';
+import { createCatchHandler } from '@/composables/service-helpers';
+
+const props = defineProps<{
+  taskKey: string,
+}>();
+
+const inputRef = ref<HTMLInputElement | null>(null);
+const query = ref<string>();
+const results = ref<ZodJiraUserPickerType[]>([]);
+
+const {
+  contributors,
+  isLoadingContributors,
+  searchContributors,
+  loadJiraContributors,
+  saveContributors,
+} = useJiraContributors(props.taskKey);
+
+onBeforeMount(loadJiraContributors);
+
+watchDebounced(query, async () => {
+  if (!isDefined(query)) {
+    return;
+  }
+  const users = await searchContributors(query.value);
+  results.value = users ?? [];
+}, { debounce: 500 });
+
+
+const onFocus = () => {
+  inputRef.value?.focus();
+};
+
+const onBlur = (event: FocusEvent) => {
+  if (
+    event.relatedTarget !== event.currentTarget
+    && !(event.currentTarget as Node)?.contains(event.relatedTarget as Node)
+  ) {
+    inputRef.value?.blur();
+    results.value = [];
+    query.value = '';
+    saveContributors().catch(createCatchHandler('Failed to save contributors', false));
+  }
+};
+
+const add = (contributor: ZodJiraUserPickerType) => {
+  contributors.value.push(contributor);
+  query.value = '';
+  results.value = [];
+};
+
+const remove = (index: number) => {
+  contributors.value.splice(index, 1);
+};
+
+</script>
+
+<template>
+  <LabelDiv
+    label="Contributors"
+    tabindex="99"
+    @click.prevent="onFocus"
+    @blur.capture="onBlur"
+  >
+    <div class="dropdown form-control">
+      <ul class="ps-0 mb-0 list-unstyled">
+        <li v-for="(contributor, index) in contributors" :key="contributor.name" class="badge text-bg-secondary">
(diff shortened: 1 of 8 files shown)


src/components/FlawContributors.vue
```diff
@@ -0,0 +1,166 @@
...
+const query = ref<string>();
+const results = ref<ZodJiraUserPickerType[]>([]);
+
+const {
+  contributors,
+  isLoadingContributors,
+  searchContributors,
+  loadJiraContributors,
+  saveContributors,
+} = useJiraContributors(props.taskKey);
```
C-Valen, line 24: Annotation for general discussion. Should non OSIDB stuff that is present on the FlawView be part of `FlawModel?

I had this exact question when worked on `internalComments`: initially made an individual composable, and at the end changed opinion and made them part of the flaw model with an unified `FlawCommentsModel` that handles both OSIDB and Jira comments ([PR](https://github.com/RedHatProductSecurity/osim/pull/285) currently on review)

My personal opinion is to make part of `FlawModel` every data we associate to a Flaw in the view, independently from the source. Reason is cause I think that way can help keeping things more organized in the frontend.

src/components/FlawContributors.vue
```diff
@@ -0,0 +1,166 @@
...
+  results.value = users ?? [];
+}, { debounce: 500 });
+
+
+const onFocus = () => {
+  inputRef.value?.focus();
+};
+
+
+const onBlur = (event: FocusEvent) => {
```
C-Valen, line 42: Minimum relevance: should we avoid double empty lines unless necessary?
May be a new lint rule candidate?

src/components/FlawContributors.vue
```diff
@@ -0,0 +1,166 @@
...
+
+      <div v-if="results.length > 0" class="menu">
+        <div
+          v-for="contributor in results"
+          :key="contributor.name"
+          class="item"
+          @click="add(contributor)"
+        >
+          <!--eslint-disable-next-line vue/no-v-html-->
+          <span v-html="sanitize(contributor.html)" />
```
C-Valen, line 107: I think we decided to let this specific warns be present as a reminder for security concerns.

src/composables/useJiraContributors.ts
```diff
@@ -0,0 +1,63 @@
+import { ref } from 'vue';
+import type { ZodJiraContributorType } from '@/types/zodJira';
+import { getJiraIssue, putJiraIssue, searchJiraUsers } from '@/services/JiraService';
+import { createCatchHandler } from './service-helpers';
+import { useMemoize, watchIgnorable } from '@vueuse/core';
+
+export default function useJiraContributors(task_key: string) {
```
C-Valen, line 7: Small preference for a more specific name, as JiraContributors doesn't give all the information about the context (contributors of a Flaw).
Also I'd aim to use as base the component's name that directly interacts with this composable (e.g. `useFlawContributors` or `useFlawJiraContributors`)

src/types/zodJira.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from 'zod';
+
```
C-Valen, line 2: Consideration of this file directly related to the using/not using FlawModel for non-OSIDB stuff.

C-Valen (commented):
I think there are some general concerns that would be interesting to discuss in a huddle.

As an additional point on the UI I think it could be preferable to use usernames instead full names as they are shorter and will make the component more compact.
However I think the username for some Jira users is the email so that'd be a problem. I was thinking on the internal comments that we maybe can generate our own usernames taken from the first part of the email of the jira user (with is a constant field) but being that the case I'd leave it as it is now.

MrMarble:
> I think there are some general concerns that would be interesting to discuss in a huddle.

We can discuss them on tomorrow's code review or make an ad-hoc call if you want

src/components/FlawContributors.vue
superbuggy: This would be my preference

src/components/FlawContributors.vue
d10n: Originally I expected OSIDB to supply all the data, but now that OSIM does some of its own data aggregation, a nice idea for the future would be to have separate OsidbFlawModel which just has OSIDB data and OsimFlawModel which contains all OSIM's aggregated data for a FlawModel

d10n (approved):

superbuggy (approved):

tboyetteredhat (approved):

PR #307, opened 2024-06-21
Task OSIDB-3002: Disable private comment creation

C-Valen:
# [OSIDB-3002] [Disable private comments creation until it is supported by OSIDB]

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [-] Test cases not added/updated
- [x] Jira ticket updated

## Summary:

Temporary disables private comment creation until it is supported from OSIDB side.

## Changes:

- Disable comment creation button on private comments tab
- Add a tooltip to inform users that the action is temporary disabled disabled
- Hide comment creation input text area on private comments tab

## Considerations:

- Didn't add reasons on the changelog as we are providing the ticket where all the info can be found

### DIFF
--- a/src/views/FlawEditView.vue
+++ b/src/views/FlawEditView.vue
@@ -41,9 +41,9 @@ function refreshFlaw() {
   <main>
     <FlawForm
       v-if="flaw"
+      :key="`${flaw.uuid}-${flaw.updated_dt}`"
       v-model:flaw="flaw"
       mode="edit"
-      :key="`${flaw.uuid}-${flaw.updated_dt}`"
       @refresh:flaw="refreshFlaw"
     />
     <div v-if="errorLoadingFlaw">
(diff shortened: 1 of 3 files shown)


src/views/FlawEditView.vue
```diff
@@ -41,9 +41,9 @@ function refreshFlaw() {
   <main>
     <FlawForm
       v-if="flaw"
+      :key="`${flaw.uuid}-${flaw.updated_dt}`"
```
MrMarble, line 44: Why this change?
It's only the order of the props?

d10n (approved):

src/views/FlawEditView.vue
C-Valen: Yes, is only the attribute order as it appeared to me on linting warns.

JakubFrejlach (approved):

PR #308, opened 2024-06-21
🐛 Prevent unhelpful error messages

superbuggy:
📝 Update changelog

# OSIDB-2962 OSIM: `[object object]` error on updating affects on the flaw
# OSIDB-2998 Error creating flaw `[object Object]` but flaw created

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Prevents `[object Object]` errors.

## Changes:

- Missing return statement (credit to @MrMarble for finding this) https://github.com/RedHatProductSecurity/osim/blob/c0dbab377091c3112b6bb65e187ab6914a56fe48/src/services/OsidbAuthService.ts#L181
- Separation of error handling into dedicated file
- Test coverage for error parsing functions
- Added return type declarations for error parsing functions

## Considerations:


tboyetteredhat (approved):

src/services/__tests__/osidb-errors-helpers.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it('should return a string with the error message when error.request is **NOT** an object', () => {
+    const error = {
+      request: 'Internal Server Error',
+    };
+    expect(getDisplayedOsidbError(error)).toBe('Internal Server Error');
+  });
+});
+
+describe('parseOsidbErrors', () => {
```
MrMarble, line 51: If we have multiple `describe` blocks, I think having a parent `describe` with the name of the module helps to organize tests. 
Not necessary for these tests, but also allows you to have a single `beforeEach` block for all the child `describe`

```js
describe('OSIDBErrorHelpers', ()=> {
	// beforeEach(()=> {})

	describe('getDisplayedOsidbError', () => {
		// ...
	})
	describe('parseOsidbErrors', () => {
		// ...
	})
});
```

src/services/osidb-errors-helpers.ts
```diff
@@ -0,0 +1,52 @@
+
+
+export function getDisplayedOsidbError(error: any): string {
+  if (error.response) {
+    if (error.response.headers?.['content-type']?.startsWith('text/html')) { // OSIDB Server might be in debug mode?
+      return parseOsidbHtmlError(error);
```
MrMarble, line 6: This case is not being tested

```js
it('should handle stack trace errors', () => {
  const error = {
    response: {
      headers: {
        'content-type': 'text/html'
      },
      data: '<div class="exception_value">Some error on line 1</div>'
    },
  };

  expect(getDisplayedOsidbError(error)).toBe(
    'Likely error between OSIDB and database:\nSome error on line 1'
  );
});
```

src/services/__tests__/osidb-errors-helpers.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it('should return a string with the error message when error.request is an object', () => {
+    const error = {
+      request: { error: 'Internal Server Error' },
+    };
+    const expectJsonStringified = '{\n  "error": "Internal Server Error"\n}';
+    expect(getDisplayedOsidbError(error)).toBe(expectJsonStringified);
+  });
+
+  it('should return a string with the error message when error.request is **NOT** an object', () => {
```
MrMarble, line 43: Not really a suggestion, just wanted to expose it in case you don't know.
I think that for this kind of tests, **tableDriven tests** ([vitest docs](https://vitest.dev/api/#test-each)) make a lot of sense, for example:

```js
it.each([
  {
    error: {/* omitted for simplicity */ },
    expected: 'OSIDB responded with error 500 (Internal Server Error). \nInternal Server Error',
    return: 'error.response.data is a string'
  },
  {
    error: {/* omitted for simplicity */ },
    expected: 'OSIDB responded with error 500 (Internal Server Error). \nerror: Internal Server Error',
    return: 'error.response.data is an object'
  },
  {
    error: {/* omitted for simplicity */ },
    expected: '{\n "error": "Internal Server Error"\n}',
    return: 'error.request is an object'
  }
])('should return a string with the error message when $return', ({ error, expected }) => {
  expect(getDisplayedOsidbError(error)).toBe(expected)
})
```

MrMarble (changes requested):
I think we should add the missing test case for `parseOsidbHtmlError`, as it was the reason the error happened on the first place

C-Valen (approved):
Good idea on making an independent error handling helpers

src/services/osidb-errors-helpers.ts
superbuggy: Can you clarify what you mean? The test does exactly what it should. In any case, I think I will alter the test case language to be more descriptive.

MrMarble: Sorry I meant in the case of a error with `text/html` `content-type`, the call to the `parseOsidbHtmlError` si not texted, the code snippet was my suggestion for the test

superbuggy: Oh yes, we should add this test case!

src/services/__tests__/osidb-errors-helpers.spec.ts
superbuggy: This is great, thanks for sharing the knowledge!

src/services/osidb-errors-helpers.ts
superbuggy: Sorry I thought the test case that **you added here** was already a part of the test suite in the PR 🫠. Thank you for the test case

src/services/__tests__/osidb-errors-helpers.spec.ts
superbuggy: Ah, good call

superbuggy:
> I think we should add the missing test case for `parseOsidbHtmlError`, as it was the reason the error happened on the first place

I think the original error was in the caller function not returning the result of the `parseOsidbHtmlError` call and we're testing it indirectly now with your suggestion above by testing `getDisplayedOsidbError`, do you think that is sufficient?

MrMarble (approved):

PR #315, opened 2024-06-25
✨ Provide label for unembargoed flaws

superbuggy:
# OSIDB-2937 Confusing Embargo badges

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [ ] Changelog updated
- [x] Jira ticket updated

## Summary:
image

## Changes:

- Provide badge for unembargoed flaws 
- Prevent edge case embargo label showing when flaw isn't embargoed (invalid data)

## Considerations:

### DIFF
--- a/src/components/IssueQueueItem.vue
+++ b/src/components/IssueQueueItem.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import { DateTime } from 'luxon';
 import { computed } from 'vue';
 
 const props = defineProps<{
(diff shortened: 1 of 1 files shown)


src/components/IssueQueueItem.vue
```diff
@@ -13,15 +13,20 @@ const props = defineProps<{
...
 const isEmbargoedButNotScheduled = computed(
-  () => props.issue.embargoed && !props.issue.unembargo_dt,
+  () => props.issue.embargoed && !props.issue.unembargo_dt
 );
 
 const isEmbargoed = computed(
-  () => isUnembargoDateScheduledForLater.value || isEmbargoedButNotScheduled.value,
+  () => (isUnembargoDateScheduledForLater.value && props.issue.embargoed)
+    || isEmbargoedButNotScheduled.value
+);
```
JakubFrejlach, line 26: I am just thinking, why does this logic needs to be so complicated?
`Flaw.embargoed == true` -> embargoed 
`Flaw.embargoed == false` -> unembargoed (public)

Or am I missing something?

JakubFrejlach (changes requested):
I am just thinking, why does this logic needs to be so complicated?
`Flaw.embargoed == true` -> embargoed 
`Flaw.embargoed == false` -> unembargoed (public)

Or am I missing something? I don't think anything with the dates is not needed for the logic.

src/components/IssueQueueItem.vue
```diff
@@ -13,15 +13,20 @@ const props = defineProps<{
...
+
+const isUnembargoed = computed(() => isEmbargoed.value
+ && DateTime.fromISO(props.issue.unembargo_dt).diffNow().milliseconds <= 0
```
superbuggy, line 29:
```suggestion
const isUnembargoed = computed(() => DateTime.fromISO(props.issue.unembargo_dt).diffNow().milliseconds <= 0
```

superbuggy:
> I am just thinking, why does this logic needs to be so complicated? `Flaw.embargoed == true` -> embargoed `Flaw.embargoed == false` -> unembargoed (public)
> 
> Or am I missing something? I don't think anything with the dates is not needed for the logic.

I'm not really sure about this PR, tbh--it stemmed from a presumably transient data integrity issue specific to one environment. Do you think there is a need for an Unembargoed label?

JakubFrejlach:
> > I am just thinking, why does this logic needs to be so complicated? `Flaw.embargoed == true` -> embargoed `Flaw.embargoed == false` -> unembargoed (public)
> > Or am I missing something? I don't think anything with the dates is not needed for the logic.
> 
> I'm not really sure about this PR, tbh--it stemmed from a presumably transient data integrity issue specific to one environment. Do you think there is a need for an Unembargoed label?

Can you please elaborate a little bit? I don't quite follow about the integrity issue. Also I don't think the label unembargoed is really needed, every Flaw is either emabrgoed or unembargoed, so either embargoed label or no label should be enough, right?

superbuggy:
Okay, if that's the single source of truth then yes. Upadted lmk if it looks good @JakubFrejlach

MrMarble (approved):

d10n (approved):

PR #321, opened 2024-06-28
OSIDB-3004: Add suggestions to flaw owner field

MrMarble:
# OSIDB-3004: Add suggestions to flaw owner field
## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Added a dropdown suggestion integrated with jira

image

Closes OSIDB-3004

### DIFF
--- a/src/components/widgets/DropDown.vue
+++ b/src/components/widgets/DropDown.vue
@@ -1,0 +1,49 @@
+<script setup lang="ts">
+</script>
+
+<template>
+  <div class="menu dropdown-menu">
+    <slot />
+  </div>
+</template>
+
+<style lang="scss" scoped>
+.menu {
+  position: absolute;
+  top: 100%;
+  right: 0;
+  z-index: 1000;
+  display: block !important;
+  color: #212529;
+  background-color: #fff;
+  border: 1px solid rgb(0 0 0 / 15%);
+  border-radius: 0.25rem;
+  max-height: 400px;
+  overflow-y: auto;
+  box-shadow: 0 0.5rem 1rem rgb(0 0 0 / 15%);
+  width: 100%;
+
+  > * {
+    display: block;
+    width: 100%;
+    padding: 0.25rem 1.5rem;
+    clear: both;
+    white-space: nowrap;
+    cursor: pointer;
+    font-size: .85rem;
+
+    &:hover {
+      background-color: #dee2e6;
+    }
+  }
+
+  :deep(.item) {
+    white-space: break-spaces;
+    display: block;
+
+    * {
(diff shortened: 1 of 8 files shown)


src/components/widgets/EditableTextWithSuggestions.vue
```diff
@@ -0,0 +1,206 @@
...
+        >{{ modelValue === '' ? placeholder : modelValue }}</span>
+        <!--if a button is inside a label, clicking the label clicks the button?-->
+        <button
+          v-if="!readOnly"
+          type="button"
+          class="osim-editable-text-pen input-group-text"
+          tabindex="-1"
+          @click="beginEdit"
+        ><i class="bi bi-pencil"></i></button>
+        <slot name="buttons-out-of-editing-mode" v-bind="{ commit, onBlur }"></slot>
```
C-Valen, line 106: I don't understand why we do require two different slots `buttons-out-of-editing-mode` and `buttons-in-editing-mode`

C-Valen (commented):
Small detail: 
- Self assign button border radius change when the input is focused

Possible UI/UX improvements:
- Add some feedback to the user always that a fetch is being performed (e.g. hiding previous results and showing a loading spinner)
- Limit suggestion box width to the field's width, currently the width is based on the longest suggested item (and may be truncating very long full names). This would be less relevant after you rebase over the new wider form style, but I think is still recommendable for UI consistency.
- For some users there are two options, I think we could filter those cases to use only the one that has an email associated

src/components/widgets/EditableTextWithSuggestions.vue
MrMarble: It was like that, is part of the `EditableText` widget, I copied the code to add the suggestions part so the API is the same on both

MrMarble:
> For some users there are two options, I think we could filter those cases to use only the one that has an email associated

I think that is only on stage, in production there's only 1 user, also we cannot filter, because not all users have the email as the user id (mine for example)

src/components/widgets/DropDown.vue
```diff
@@ -0,0 +1,42 @@
+<script setup lang="ts">
+defineProps<{
+  show: boolean,
+}>();
```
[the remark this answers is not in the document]
MrMarble: Yeah that is easier 😆

C-Valen (approved):

tboyetteredhat (approved):

superbuggy (approved):

PR #328, opened 2024-07-02
Add test case for filed trackers list

huiwangredhat:
# [OSIDB-3007] [List Filed trackers]

Set up: export FLAW_ID=$CVE_id and the flaw has affect.


RichardXuan (approved):
LGTM

PR #330, opened 2024-07-02
Enforce jira link on pull requests

MrMarble:
This PR adds a new github action that enforces the use of `Closes OSIDB-1234` or `Fixes OSIDB-1234` for linking the PR with Jira.

Added a comment to the PR template that reminds the user to add the tag as well.

In case that the PR does not have a Jira task associated, add the `internal` label to skip the check


JakubFrejlach (approved):
LGTM! Nice!

PR #343, opened 2024-07-08
Allow deleting CVSS scores

dmonzonis:
# OSIDB-1843 Allow deleting CVSS scores

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Right now it is not possible to delete a flaw's CVSS score through OSIM. Even though there is a button to clear the field, when trying to save the flaw it will show an error saying that this field cannot be null.

With this commit, when clearing the CVSS score field and saving the flaw, it will delete the CVSS score through OSIDB's API, using the DELETE method.

Closes OSIDB-1843.

### DIFF
--- a/src/composables/useFlawCvssScores.ts
+++ b/src/composables/useFlawCvssScores.ts
@@ -1,4 +1,4 @@
-import { putFlawCvssScores, postFlawCvssScores } from '@/services/FlawService';
+import { deleteFlawCvssScores, putFlawCvssScores, postFlawCvssScores } from '@/services/FlawService';
 import { computed, ref, watch, type Ref } from 'vue';
 import type { ZodFlawType } from '@/types/zodFlaw';
 import { groupWith, equals } from 'ramda';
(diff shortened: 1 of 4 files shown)


src/composables/useFlawCvssScores.ts
```diff
@@ -92,9 +92,14 @@ export function useFlawCvssScores(flaw: Ref<ZodFlawType>) {
 
   async function saveCvssScores() {
     if (flawRhCvss3.value?.created_dt) {
+      // Handle existing CVSS score
+      if (flawRhCvss3.value?.vector === null && flawRhCvss3.value?.uuid !== null) {
```
superbuggy, line 96:
```suggestion
      if (flawRhCvss3.value?.vector === null && flawRhCvss3.value?.uuid != null) {
```

superbuggy: `!=` will allow undefined, which likely would be the value if the cvss data were not saved

C-Valen (approved):

superbuggy (changes requested):

MrMarble (approved):

superbuggy (approved):

PR #345, opened 2024-07-09
OSIDB-3123: Invalid Affect Resolution

MrMarble:
# OSIDB-3123: Invalid Affect Resolution

## Checklist:

- [x] Linting passed
- [x] Type checks passed
- [x] Tests suite passed
- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

When the `affectedness` was changed, the `resolution` kept the previous value, only the available options where updated but not the affect model.

## Changes:

* Reset the `resolution` value when the `affectedness` is changed
* Simplify the logic by using the available `affectedness` and `resolution` enums
* Changed the empty value (`''`) to an `Empty` string so it's easier to differentiate empty vs undefined
* Added a validation to the `resolution` based on the possibles values. See https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/constants.py#L61-L76
## Considerations:

Closes OSIDB-3123

https://github.com/RedHatProductSecurity/osim/assets/4268580/c483d8ee-b05b-4c5f-a714-6616bf6d057e

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Duplicated loading spinner on flaw lists (`OSIDB-3092`)
 * Internal comments creation fails on chrome browser (`OSIDB-3091`)
 * Auto commit edited references and ackowledgements when start editting a new one (`OSIDB-2905`)
+* Affects resolution is not updated after changing affectedness (`OSIDB-3123`)
 
 ## [2024.7.0]
 ### Changed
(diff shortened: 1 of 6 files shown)


superbuggy:
Should we italicize the Empty choice in the dropdown?

superbuggy (approved):
Looks solid

C-Valen (approved):

tboyetteredhat (approved):

MrMarble:
> Should we italicize the Empty choice in the dropdown?

I tried but seems that the `option` html element does not likes styles, at least firefox ignored it. We would need to use a custom dropdown element to handle styles

PR #348, opened 2024-07-11
[features] Add flaw state transition scenario

qduanmu:
Closes OSIDB-3063

### DIFF
--- a/features/flaw_detail_public.feature
+++ b/features/flaw_detail_public.feature
@@ -151,3 +151,14 @@ Feature: Flaw detail testing on public flaw
     Scenario: Filter unfiled trackers
       When I add some affects with valid data
       Then I could filter trackers by stream or component name
+
+    Scenario Outline: Update flaw state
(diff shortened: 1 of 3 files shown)


huiwangredhat (commented):
I tried 3 or 4 times and failed. The errors:
self.check_element_exists(By.XPATH, f'//*[contains(text(), "{value}")]')
        File "/Users/huiwang/348/osim/features/pages/base.py", line 57, in check_element_exists
          return WebDriverWait(self.driver, self.timeout).until(
raise ElementNotFoundException(
      seleniumpagefactory.Pagefactory.ElementNotFoundException: An exception of type TimeoutException occurred. With Element -: flawPromotedMsg - locator: (xpath, //strong[text()='Flaw Promoted'])

huiwangredhat (approved):
LGTM except the timeout issue as we discussed in the meeting. Later if the timeout is needed to be updated on the pipeline, let's update it.

PR #350, opened 2024-07-12
fix failed test cases and add script to execute feature files by order

RichardXuan:
Fixes OSIDB-3043


run_automation_test.sh
MrMarble, line 1: Once merged I'll change the Pipeline to run this script instead of `behave`

RichardXuan: Sure. Thanks. I noticed there is a folder called scripts https://github.com/RedHatProductSecurity/osim/tree/main/scripts, should i place this script to this folder. And i notice the dir when executing `behave` cmd is https://gitlab.corp.redhat.com/product-security/dev/osim-ops/-/blob/new-pipeline/roles/pipeline-project/k8s/acceptance-test.yml.j2#L51 , is this the root dir of osim code repo.

MrMarble: Sure, I think it makes sense to put the script in the `scripts` folder, so it can be run with `./scripts/run_automation_test.sh`
The root dir is the OSIM repository, where you normally run `behave` command, like `behave features/login.feature`.

RichardXuan: Ok. Thanks.

huiwangredhat (commented):
Step: And I create new embargoed flaw with valid data  
 Error:  seleniumpagefactory.Pagefactory.ElementNotFoundException: An exception of type TimeoutException occurred. With Element -: comment#0Text - locator: (xpath, //span[text()='Comment#0'])

huiwangredhat (commented):
except the affects' and trackers' related test cases and one reference case, the others works well.

huiwangredhat (approved):
After the second round running, there is no the error of reference case. After discussing, we found the root cause is the network timeout.

PR #358, opened 2024-07-22
Feature OSIDB-3138: Search non empty CVE description

C-Valen:
# OSIDB-3138: Search non empty CVE description

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Context:

Due to a bug there are lot of flaws with empty CVE Description on Requested status, which difficult the work of users as it is generally expected to receive flaws with non empty description when looking for status Requested.

## Summary:

Adds support for searching flaws with non empty CVE description on `AdvancedSearch` view as a workaround for users search queries. It is provided a checkbox on Advanced search which modifies the CVE Description Field setting the value to `nonempty` which triggers the param for the OSIDB request.

## Changes:

- Adds a checkbox `Non empty CVE description` on advanced
- Supports `nonempty` value for CVE Descriptions params

## Demo

image

https://github.com/user-attachments/assets/60a0e04c-fc48-4d3b-a109-7ab1a592213e

## Considerations:

- There probably are cleaner ways to do this (both for code and UI), but as it is a temporary workaround my aim was to cover the need in the simplest and quickest way.
- I had to put the `shouldShowAdvanced` ref initialized to `true` cause the advanced search was always loaded in hidden state.

Closes OSIDB-3138

### DIFF
--- a/src/components/IssueSearchAdvanced.vue
+++ b/src/components/IssueSearchAdvanced.vue
@@ -17,6 +17,8 @@ const emit = defineEmits(['filter:save']);
 
 const isNonEmptyDescriptionSelected = ref(false);
 
+const descriptionParamValue = computed(() => facets.value.find(facet => facet.field === 'cve_description')?.value);
+
 watch(isNonEmptyDescriptionSelected, () => {
   const facet = facets.value.find(facet => facet.field === 'cve_description');
   if (isNonEmptyDescriptionSelected.value) {
@@ -25,8 +27,17 @@ watch(isNonEmptyDescriptionSelected, () => {
     } else {
       facet.value = 'nonempty';
     }
-  } else if (facet) {
+  } else if (facet && descriptionParamValue.value === 'nonempty') {
     facet.value = '';
+  }
+});
+
+watch(descriptionParamValue, (value) => {
+  if (value !== 'nonempty' && isNonEmptyDescriptionSelected.value) {
+    isNonEmptyDescriptionSelected.value = false;
+  }
+  if (value === 'nonempty' && !isNonEmptyDescriptionSelected.value) {
+    isNonEmptyDescriptionSelected.value = true;
   }
 });
 
--- a/src/components/IssueSearchAdvanced.vue
+++ b/src/components/IssueSearchAdvanced.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
-import { computed, ref } from 'vue';
+import { computed, ref, watch } from 'vue';
 import { flawImpacts, flawSources, flawIncidentStates } from '@/types/zodFlaw';
-import { useRoute } from 'vue-router';
+import LabelCheckbox from '@/components/widgets/LabelCheckbox.vue';
 import { flawFields } from '@/constants/flawFields';
 import { useSearchParams } from '@/composables/useSearchParams';
 import { descriptionRequiredStates } from '@/types/zodFlaw';
@@ -16,6 +16,21 @@ const props = defineProps<{
 const route = useRoute();
 
 const emit = defineEmits(['filter:save']);
+
+const isNonEmptyDescriptionSelected = ref(false);
+
+watch(isNonEmptyDescriptionSelected, () => {
+  const facet = facets.value.find(facet => facet.field === 'cve_description');
+  if (isNonEmptyDescriptionSelected.value) {
+    if (!facet) {
+      facets.value.push({ field: 'cve_description', value: 'nonempty' });
+    } else {
+      facet.value = 'nonempty';
+    }
+  } else if (facet) {
+    facet.value = '';
+  }
+});
 
 const nameForOption = (fieldName: string) => {
   const mappings: Record<string, string> = {
@@ -123,6 +138,11 @@ const shouldShowAdvanced = ref(route.query.mode === 'advanced');
       >
         Save as Default
       </button>
+      <LabelCheckbox
+        v-model="isNonEmptyDescriptionSelected"
+        label="Non Empty CVE Description"
+        class="d-inline-block"
+      />
     </form>
   </details>
 </template>
--- a/src/components/IssueSearchAdvanced.vue
+++ b/src/components/IssueSearchAdvanced.vue
@@ -13,8 +13,6 @@ const props = defineProps<{
   isLoading: boolean;
 }>();
 
-const route = useRoute();
-
 const emit = defineEmits(['filter:save']);
 
 const isNonEmptyDescriptionSelected = ref(false);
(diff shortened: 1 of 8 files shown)


src/components/IssueSearchAdvanced.vue
```diff
@@ -71,7 +84,7 @@ const optionsFor = (field: string) =>
     major_incident_state: flawIncidentStates,
     affects__affectedness: affectAffectedness,
   })[field] || null;
-const shouldShowAdvanced = ref(route.query.mode === 'advanced');
+const shouldShowAdvanced = ref(true);
```
MrMarble, line 87: If the `mode` query parameter is not used anymore, I'll remove it, is used in multiple files https://github.com/search?q=repo%3ARedHatProductSecurity%2Fosim+%22%3A+%27advanced%27%22&type=code

MrMarble:
Another option could be adding the checkbox inside the `v-foor` loop that builds the search form, using the `allowedEmptyFieldMapping` as a filter, so we support searching for nonempty values on all fields, instead of only one, and because the field needs to be added to the form for the check to be visible, it will not jump when checking it

src/components/IssueSearchAdvanced.vue
```diff
@@ -13,10 +13,23 @@ const props = defineProps<{
...
+  if (isNonEmptyDescriptionSelected.value) {
+    if (!facet) {
+      facets.value.push({ field: 'cve_description', value: 'nonempty' });
+    } else {
+      facet.value = 'nonempty';
+    }
+  } else if (facet) {
+    facet.value = '';
+  }
+});
```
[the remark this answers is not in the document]
C-Valen: Thanks for the feedback @MrMarble 

I replicated the problem you mentioned. May be we can use another watcher on the input field and changing the checkbox value of checked? Wdyt?

src/composables/useFlawsFetching.ts
```diff
@@ -13,8 +13,8 @@ export function useFlawsFetching() {
   function finializeRequestParams(params: any = {}){
     const requestParams: Record<string, any> = {};
     for (const key in params) {
-      if (params[key] === '' && allowedEmptyFieldMapping[key]) {
-        requestParams[allowedEmptyFieldMapping[key]] = 1;
+      if (['', 'nonempty'].includes(params[key]) && allowedEmptyFieldMapping[key]) {
```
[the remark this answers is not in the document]
C-Valen: IMO values that explicitly provide the meaning to the users.are preferable.
Empty string and * I think can be more confusing to new users.

This is why I used 'nonempty'. However as this is PR a temporary workaround I didn't want to change the implementation for empty strings.

Good consideration for team discussion as @tboyetteredhat is working on the redesign of the advanced search.

src/components/IssueSearchAdvanced.vue
C-Valen: I am not sure if that's something is/will be needed, may be @tboyetteredhat of @superbuggy can confirm

src/components/IssueSearchAdvanced.vue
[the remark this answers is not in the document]
MrMarble: Maybe we can use a combination of the `@change` event to trigger the updates (the code that is inside the watch now) and changing the `isNonEmptyDescriptionSelected` to a computed that reads the state:

```js
const isNonEmptyDescriptionSelected = computed(()=> 
  facets.value.find(facet => facet.field === 'cve_description')?.value === 'nonempty'
);

const toggleEmptyCVEDescription = (event: Event) => {
  const facet = facets.value.find(facet => facet.field === 'cve_description');
  const checked = (event.target as HTMLInputElement).checked

  if (checked) {
    if (!facet) {
      facets.value.push({ field: 'cve_description', value: 'nonempty' });
    } else {
      facet.value = 'nonempty';
    }
  } else if (facet) {
    facet.value = '';
  }
}:
```

```vue
<LabelCheckbox
        :checked="isNonEmptyDescriptionSelected"
        @changed="toggleEmptyCVEDescription"
        label="Non Empty CVE Description"
        class="d-inline-block"
      />
```

src/components/IssueSearchAdvanced.vue
tboyetteredhat: We are using mode to show advanced search section.

MrMarble: @tboyetteredhat The only usage I see is this one that @C-Valen removed, where is used apart from this file?

src/components/IssueSearchAdvanced.vue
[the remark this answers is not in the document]
C-Valen: Hey @MrMarble I was trying your suggested approach but I got into some problems with it cause of circular cycles on the input/checkbox events. At the end I decided to add a second watcher on the input `CVE Description field`.

src/components/IssueSearchAdvanced.vue
MrMarble: If is not read anywhere I think we could remove it

src/components/IssueSearchAdvanced.vue
```diff
@@ -13,10 +13,34 @@ const props = defineProps<{
   isLoading: boolean;
 }>();
 
-const route = useRoute();
-
 const emit = defineEmits(['filter:save']);
 
+const isNonEmptyDescriptionSelected = ref(false);
+
+const descriptionParamValue = computed(() => facets.value.find(facet => facet.field === 'cve_description')?.value);
```
MrMarble, line 20: Saving the entire `facet` instead of the value you can reuse it also on line 23
```suggestion
const descriptionParamValue = computed(() => facets.value.find(facet => facet.field === 'cve_description'));
```

src/components/IssueSearchAdvanced.vue
```diff
@@ -13,10 +13,34 @@ const props = defineProps<{
...
+
+watch(isNonEmptyDescriptionSelected, () => {
+  const facet = facets.value.find(facet => facet.field === 'cve_description');
```
MrMarble, line 23: This can be removed and use computed `descriptionParamValue`

src/components/IssueSearchAdvanced.vue
C-Valen: I thought the same but on the watcher (line 23) I need also to update the value in some cases so I need a writable reference.

MrMarble: Well you could use the index

MrMarble (approved):

tboyetteredhat (approved):

PR #364, opened 2024-07-24
Add test case for advance search with cve description and cve descrip…

huiwangredhat:
Fixes OSIDB-3172
Requirement: export FLAW_ID=$flaw_id which has the not empty cve description and review status.


RichardXuan (approved):
LGTM

PR #365, opened 2024-07-24
🐛 Fix: Regression UI bug on textarea descriptions layout

C-Valen:
# OSIDB-3174: Style for Comment#0 reverted back to previous version

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Test cases added/updated
- [x] Jira ticket updated

## Summary:

Fix regression issue on text-area description fields and the focus highlight - label overlap

## Changes:

- Fix inheritable classes on text-area widget
- Fix text-area focus highlight label overlap

Closes OSIDB-3174

PR #369, opened 2024-07-29
OSIDB-3189: Filter jira users by task key

MrMarble:
# OSIDB-3189: Filter jira users by task key

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Filter out jira users that does not have access to the current project.

The endpoint `api/2/user/picker` does not allow to specify a task ID and returns all the jira users, currently there isn't an available endpoint from the public API that allows this kind of filtering while maintaining  the option to query by any part of the user (name, username or email).
I ended up using an internal endpoint `/internal/2/users/assignee` that allows filtering via task id and querying any part of the user (name, username or email)

### Snapshot testing

I took the opportunity to do a small proof of concept about snapshot testing, so we can start using it from now on.
This kind of tests will prevent us for doing regressions like in #365 as they match the whole html tree of the component, so any change in classes, attributes or structure will make the test **fail**.

This will also simplify testing, instead of manually expecting multiple values to `exist` or to have an specific value.
As an example, all this code:
https://github.com/RedHatProductSecurity/osim/blob/c1683742d1eb032ec5aaf54c275f86cfa3829e39/src/components/__tests__/CvssNISTForm.spec.ts#L89-L97
can be replaced with:
```js
expect(wrapper.html()).toMatchSnapshot();
```

Here's the link to the [test file](https://github.com/RedHatProductSecurity/osim/blob/fix/OSIDB-3189-jira-user-search/src/components/widgets/__tests__/JiraUser.spec.ts) and the generated [snapshot](https://github.com/RedHatProductSecurity/osim/blob/fix/OSIDB-3189-jira-user-search/src/components/widgets/__tests__/__snapshots__/JiraUser.spec.ts.snap)

Closes OSIDB-3189

### DIFF
--- a/src/components/widgets/JiraUser.vue
+++ b/src/components/widgets/JiraUser.vue
@@ -1,0 +1,29 @@
+<script setup lang="ts">
+import { computed, toValue } from 'vue';
+
+const props = defineProps<{
+  displayName: string;
+  emailAddress?: string;
+  name?: string;
+  query?: string;
+}>();
+
+const formatedUser = computed(() => {
+  let output = props.displayName;
+  if (props.emailAddress) {
+    output += ` - ${props.emailAddress}`;
+  }
+  if (props.name) {
+    output += ` (${props.name})`;
+  }
+  if (props.query) {
+    output = output.replace(new RegExp(toValue(props.query), 'gi'), '<b>$&</b>');
+  }
+  return output;
+});
+</script>
+
+<template>
+  <!-- eslint-disable-next-line vue/no-v-html -->
+  <span v-html="formatedUser" />
+</template>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 ### Added
 * Indicate when trackers are unexpectedly not available (`OSIDB-3158`)
 
+### Changed
+* Jira suggestions are now filtered by the current project (`OSIDB-3189`)
+
 ## [2024.7.1]
 
 ### Added
--- a/src/components/FlawComments.vue
+++ b/src/components/FlawComments.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { computed, nextTick, onMounted, ref, watch } from 'vue';
 import { isDefined, watchDebounced } from '@vueuse/core';
-import type { ZodJiraUserPickerType } from '@/types/zodJira';
+import type { ZodJiraUserAssignableType } from '@/types/zodJira';
 import LabelTextarea from '@/components/widgets/LabelTextarea.vue';
 import sanitizeHtml from 'sanitize-html';
 import { osimRuntime } from '@/stores/osimRuntime';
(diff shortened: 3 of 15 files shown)


src/components/widgets/JiraUser.vue
```diff
@@ -0,0 +1,29 @@
...
+const formatedUser = computed(() => {
+  let output = props.displayName;
+  if (props.emailAddress) {
+    output += ` - ${props.emailAddress}`;
+  }
+  if (props.name) {
+    output += ` (${props.name})`;
+  }
+  if (props.query) {
+    output = output.replace(new RegExp(toValue(props.query), 'gi'), '<b>$&</b>');
```
superbuggy, line 20: Is the use of a RegExp here in place of a string to filter out multiple occurrences?

superbuggy (approved):

tboyetteredhat (approved):

src/components/widgets/JiraUser.vue
MrMarble: Yes, and also to keep the original case of the text.
For example, given this output: `Alvaro Tinoco - atinoco@email.com (atinoco)`, the query `tinoco` will transform to `Alvaro <b>Tinoco</b> - a<b>tinoco</b>@email.com (a<b>tinoco</b>)` (note the **uppercase T** on my last name), using a string the casing will be lost

superbuggy: Nice, thank you for the context

superbuggy:
So when making an intentional UI change we would follow [this process for updating a snapshot](https://vitest.dev/guide/snapshot#updating-snapshots) and commit the snapshot, sounds like--is that right? 

Did we also want to use image snapshots for views? If so, I think that could wait for another PR.

MrMarble:
> So when making an intentional UI change we would follow [this process for updating a snapshot](https://vitest.dev/guide/snapshot#updating-snapshots) and commit the snapshot, sounds like--is that right?

Exactly, it is as easy as `vitest -u`, could be made into a yarn script like we have `yarn unit-test`, i.e `yarn unit-test-update`

> Did we also want to use image snapshots for views? If so, I think that could wait for another PR.

For visual snapshots we need to install https://github.com/americanexpress/jest-image-snapshot for the method `expect().toMatchImageSnapshot()` to be available.

Visual snapshots are great for style changes, maybe we could implement them only for the views instead of the components. 
I think for that to work properly we need to first create mocks of everything backend related, as different data on the screenshot means that it will fail to compare, also we would need to host the snapshot somewhere on the repository, and having real data (even for stage data) could be a security risk, that said I think they can also be a great addition.

We could run the tests as a Workflow, here's an [example](https://github.com/MrMarble/drg-editor/blob/develop/.github/workflows/screenshots.yml) from one of my repositories and a PR with the automated comments implying the tests failed https://github.com/MrMarble/drg-editor/pull/28.
Sadly I deleted the images from the hosted service so they can not be longer viewed but here's one screenshot I had saved of a test that failed to load the correct font

<details>
<summary>Screenshot</summary>

image

</details>

C-Valen (approved):

PR #371, opened 2024-07-29
OSIDB-2818: Affect section redesign

C-Valen:
# OSIDB-2818 Affect section redesign

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Performs the re-design of the full  `Affected Offerings` section, providing new components for Affects, Trackers and Tracker Manager views. It also covers some fixes, provides new functionalities, removes deprecated/old code and aims to improve components naming and hierarchy.

Almost all changes are UI related, but there are specific minimal changes on `FlawAffectsModel` (for having added affects as a parameter) and `useTrackers` (for `loading` functionality and improve selection toggling)

__Note!:__ The PR shouldn't be merged until approval from VM team on last changes (meeting on Thursday 08/08),
but as most of the changes have been already approved/requested by them, not many things are likely to change, so it can start being reviewed.

### General
- New component file names: `FlawAffects`, `FlawTrackers` and `TrackersManager`
- Add an `Affected Modules` filter list on the start of the section
- Renamed OSIM Setings variable from `apiKeys` to `osimSettings` and storage from `OSIM::API-KEYS` to `OSIM::USER-SETTINGS` as they are now not limited to the api keys
- Added bind attributes to `modal-dialog` on `Modal` widget to be able to apply parent classes and styles
- Provides an `usePagination` composable

### Affects View
- Change affect view to a paginated table
- Use settings store to keep user's affects per page value
- Add a toolbar to the affect table (cutomize affects per page, filter by affect states and perform multiple affect operations)
- Support affects table sorting by different fields
- Support affects field-filters for `Affectedness`, `Resolution` and `Impact` fields
- Add dynamic affect display based on it's state (new | added  | removed | editing | modified)
- Support multiple affects selection
- Support filing trackers for individual/selected affects
- Use blank affect as parameter on the add affect function
- Change blank affect values and move it from the composable to the `FlawAffects` component

### Trackers View
- Change tracker view to a paginated table
- Fix `affected module` information, now getting it from affects mapping
- Add `Bug ID` information
- Remove `Type` information
- Support sorting trackers by updated/created date
- Support trackers field-filter for `Status` field
- Add new `Resolution` column

### Trackers Manager View
- Integrated/embedded view style for `TrackersManager` within the trackers view
- Modal view for `TrackersManager` for individual/selected affects
- Change layout of `TrackersManayer` for a more compact view
- Sightly change `TrackersManager` style for a more homogeneous view
- Implement "Loading trackers" feature for better UI/UX on fetching states
- Improve style states on "all trackers" toggle selection buttons

### Tests
- Improvements and restructure on `SampleData` for more flexible and wide dummy data
- Provide new testing suite for `FlawAffects` component
- Provide new testing suite for `FlawTrackers` component
- Provide new testing suite for `TrackersManager` component

## Considerations:

- All existing test suites for old affect section and trackers have been removed as they were covering correct UI placement, I didn't add such tests for new views considering to rely on the recently implemented snapshots tests.

- I considered the rename of the 'Trackers Manager' individual section parts (Available Trackers | Selected to file | Already Filed | Untrackable Affects)

- Also considered a sightly adjustment on the `Trackers Manager` layout/UI aiming for a bit more homogeneous style in the individual parts. For this purpose the untrackable affects style have been modified and the information text have been moved into a icon tooltip to optimize the space.

- Keeping full flaw trackers manager can seem redundant as it is possible open the manager form the selection of all affects, but it was decided to keep it as it provides a different view that doesn't cover the trackers table (as the modal does) for the cases that is needed to manage trackers looking at the table at the same time

- I finally considered not adding affects/trackers per page on user settings view to wait to have more OSIM custom UI settings to add so when provided to the users it offers a more extensive settings.

- I consider it would be nice to update the scss color palettes to have a consistent set of reusable variable colors

- As a side note I noticed widget page (only visible in dev) is kind f broken, I suggest removing it unless someone is/will use it.

### Affected Modules Filters
image

### Affects Table Toolbar
image

### Affects Table
image

### Modal Tracker Manager (loading)
image

### Modal Tracker Manager (loaded)
image

### Trackers Table
image

### Embedded Trackers Manager
image

### No Affects View
image

## Demos
Changes after the demo recordings:
- Added resolution column on trackers table
- Swap Module/Component columns order on affects table

https://github.com/user-attachments/assets/b4b7a9a3-e693-494b-b6e6-ca6169f54110

https://github.com/user-attachments/assets/e1393543-1b40-425a-af95-de686f2566b7

Closes OSIDB-2818
Closes OSIDB-3076
Closes OSIDB-2789
Closes OSIDB-3022
Closes OSIDB-3250


src/components/TrackersManager.vue
```diff
@@ -0,0 +1,309 @@
...
+import { useTrackers } from '@/composables/useTrackers';
+import { computed, toRef } from 'vue';
+
+const props = defineProps<{
+  flawId: string;
+  affects: ZodAffectType[];
+  mode: 'modal' | 'embedded';
+}>();
+
+const affects = toRef(props, 'affects');
```
MrMarble, line 12: This is only used on line 26, on line 29 and 40 you use `props.affects`
I think you can remove it and also use `props.affects` on line 26

src/components/TrackersManager.vue
```diff
@@ -0,0 +1,309 @@
...
+      return refAffect.ps_component === tracker.ps_component
+      && refAffect.ps_module === tracker.ps_module;
+    }
+    );
+  });
+});
+
+const emit = defineEmits<{
+  'affects-trackers:refresh': [];
+}>();
```
MrMarble, line 52: Maybe move this to the top, next to the `defineProps`?

src/components/TrackersManager.vue
```diff
@@ -0,0 +1,309 @@
...
+
+function handleFileTrackers() {
+  fileTrackers().then(() => {
```
MrMarble, line 55: What happens if the `fileTrackers` is rejected? do we need a `.catch`?

src/components/TrackersManager.vue
```diff
@@ -0,0 +1,309 @@
...
+  <div class="trackers-manager py-3 px-4" :class="{'embedded': mode === 'embedded'}">
+    <h4 v-if="mode === 'embedded'" class="mb-2 d-flex">
+      Trackers Manager
+    </h4>
+    <div class="d-flex gap-2">
+      <input
+        v-model="filterString"
+        type="text"
+        class="form-control form-control-sm"
+        style="width: 35%;"
```
MrMarble, line 71: Do we want to use `style` attributes? or should we keep all styling inside the `style` tag?

src/components/TrackersManager.vue
```diff
@@ -0,0 +1,309 @@
...
+      <div class="col">
+        <div class="mb-2">
+          <div v-if="isLoadingTrackers" class="ms-2 mb-1">
+            <span
+              class="spinner-border spinner-border-sm"
+              role="status"
+            >
+              <span class="visually-hidden">Loading...</span>
+            </span>
+            <span> Loading trackers&hellip; </span>
```
MrMarble, line 118: why the spaces around it?
```suggestion
            <span>Loading trackers&hellip;</span>
```

I noticed you have them on all the `Loading xxx` and `No available xxx` strings, was that intended?

src/components/TrackersManager.vue
```diff
@@ -0,0 +1,309 @@
...
+      @click="handleFileTrackers"
+    >
+      <i v-if="!isFilingTrackers" class="bi bi-archive"></i>
+      File Selected Trackers
+    </button>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+@import '@/scss/bootstrap-overrides';
```
MrMarble, line 259: This is already imported in the index, do we need it here? https://github.com/RedHatProductSecurity/osim/blob/bb264759f1c7e4fc460e5d6d7e682a2974a7b325/src/scss/index.scss#L1

src/components/FlawForm.vue
```diff
@@ -1,15 +1,14 @@
 <script setup lang="ts">
 import { DateTime } from 'luxon';
 import { computed, ref, watch, onMounted } from 'vue';
-import { sortWith, ascend, prop } from 'ramda';
+// import { sortWith, ascend, prop } from 'ramda';
```
MrMarble, line 4: maybe remove it instead of commenting it out?
Also on line 26 and 206

src/components/widgets/Modal.vue
```diff
@@ -38,7 +38,7 @@ watch(show, () => {
       aria-labelledby="modalTitle"
       :role="show ? 'dialog' : ''"
     >
-      <div class="modal-dialog">
+      <div class="modal-dialog" v-bind="$attrs">
```
MrMarble, line 41: Do we want to disable inheritance as well?

```js
defineOptions({
  inheritAttrs: false
})
```

src/composables/usePagination.ts
```diff
@@ -0,0 +1,48 @@
+import { type Ref, ref, computed, watch } from 'vue';
+
+export function usePagination(totalPages: Ref<number>, initialItemsPerPage: number) {
+  const minItemsPerPage = 5;
+  const maxItemsPerPage = 20;
+  const maxPagesToShow = 7;
```
MrMarble, line 6: I would change this to be configurable but optional with default values
```suggestion
interface PaginationOptions {
  minItemsPerPage?: number,
  maxItemsPerPage?: number,
  maxPagesToShow?: number,
}

const defaultPaginationOptions: PaginationOptions = {
  minItemsPerPage: 5,
  maxItemsPerPage: 20,
  maxPagesToShow: 7,
}

export function usePagination(totalPages: Ref<number>, initialItemsPerPage: number, options?: PaginationOptions) {
  const {  minItemsPerPage,maxItemsPerPage, maxPagesToShow } = {...defaultPaginationOptions, ...options}
```

Using this patterns allows the user to set only part of the options and keep the default for the rest.
Since this is now part of the input, can be removed from the `return`

src/stores/SettingsStore.ts
```diff
@@ -44,6 +52,8 @@ export const useSettingsStore = defineStore('SettingsStore', () => {
       showNotifications: false,
       bugzillaApiKey: '',
       jiraApiKey: '',
+      affectsPerPage: 10,
+      trackersPerPage: 10,
```
MrMarble, line 56: Should we use the `defaultValues` variable for this?

```typescript
function $reset() {
  settings.value = defaultValues; // maybe {...defaultValues} in case there's some object reference
}
```

MrMarble (commented):
Not that important, but normally `__fixtures__` is the word used instead of `__sampledata__`

I'll continue the review later, I'm still missing some files

src/components/TrackersManager.vue
superbuggy: maybe use `ch` units for this field, like`25ch`

src/components/TrackersManager.vue
superbuggy: May not be needed here since no sass variables are used here

src/components/TrackersManager.vue
C-Valen: I'd say that the catching should happen on the origin method `fileTrackers` at the `useTrackers` composable.

src/composables/usePagination.ts
C-Valen: Checking this point I noticed min and max items per page are not used in the `usePagination` but on the components, as is needed only for the condition before updating the setting.

Because of that I considered removing them from the composable, and adjusting the `maxPagesToShow` to be optional parameter with default value 7.

C-Valen:
@MrMarble @superbuggy the PR has been updated addressed the commented points from the last code review.

src/components/FlawForm.vue
```diff
@@ -475,17 +431,17 @@ const theAffects = computed(() => {
...
             :affectsToDelete="affectsToDelete"
             :error="errors.affects"
             :flawId="flaw.uuid"
-            @affect:recover="(affect) => recoverAffect(flaw.affects.indexOf(affect))"
+            :embargoed="flaw.embargoed"
+            @affect:recover="(affect) => recoverAffect(affectsToDelete.indexOf(affect))"
             @affect:remove="(affect) => removeAffect(flaw.affects.indexOf(affect))"
             @affects:refresh="refreshAffects"
-            @add-blank-affect="addBlankAffect"
+            @affect:add="(affect) => addAffect(affect)"
```
MrMarble, line 444:
```suggestion
            @affect:add="addAffect"
```

src/components/FlawForm.vue
```diff
@@ -504,39 +460,28 @@ const theAffects = computed(() => {
...
+        <button
+          v-osim-loading.grow="isSaving"
+          type="submit"
+          class="btn btn-primary ms-3"
+          :disabled="isSaving"
+        >
+          Save Changes
+        </button>
+      </div>
+      <div v-if="mode === 'create'">
```
MrMarble, line 476:
```suggestion
      <div v-else-if="mode === 'create'">
```

src/composables/usePagination.ts
```diff
@@ -0,0 +1,43 @@
+import { type Ref, ref, computed, watch } from 'vue';
+
+export function usePagination(totalPages: Ref<number>, initialItemsPerPage: number, maxPagesToShow?: number) {
```
MrMarble, line 3: Setting the default value on the function declaration removes the need for the `or` on line 9, also it is shown by VSCode so you can see the default value without reading the code

https://github.com/RedHatProductSecurity/osim/blob/5eda9e9e4fe8ae821210bb9abc9aae58ec169342/src/composables/usePagination.ts#L9


```suggestion
export function usePagination(totalPages: Ref<number>, initialItemsPerPage: number, maxPagesToShow = 7) {
```

src/components/FlawAffects.vue
MrMarble, line 1: We need to define a convention for components, a component with 1.519 lines **should be avoided** at all times.
As this is the general scaffolding we use right now, I'll not ask to change this, but we have to address it some how for future work.

C-Valen: 100% agreed. I was thinking on that while working on this.
What I have thought is to use directory-based components, for example `FlawAffects` could be:

	-src/components
	... 
	--/FlawAffects
	---index.vue
	---AffectsToolbar.vue
	---/AffectsTable.vue
	----AffectsTableHeader.vue
	----AffectsTableRow
	etc.

That way we could adjust granularity per component, have more flexibility on reusing atomic components and having more maintainable and organized code.

C-Valen:
Update: Rebased and integrated last Affect Resolution changes

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,460 @@
...
+  allTrackersCount: number;
+}>();
+
+const emit = defineEmits<{
+  'file-tracker': [value: object];
+  'affects:refresh': [];
+}>();
+
+const settingsStore = useSettingsStore();
+const settings = ref(settingsStore.settings);
```
MrMarble, line 25: `settingsStore.settings` is already a reference

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,460 @@
...
+
+  const customSortFn = (affect: TrackerWithModule) => {
+    return affect[customSortKey] || 0;
+  };
+
+  const comparator = [order<TrackerWithModule>(customSortFn)];
+
+  return sortWith([
+    ...comparator
+  ])(trackers);
```
MrMarble, line 64: Since comparator is already an array and is the only item I think can be changed 
```suggestion
  const comparator = [order<TrackerWithModule>(customSortFn)];

  return sortWith(comparator)(trackers);
```

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,460 @@
...
+
+const paginatedTrackers = computed(() => {
+  const start = (currentPage.value - 1) * settings.value.trackersPerPage;
+  const end = start + settings.value.trackersPerPage;
+  return sortedTrackers.value.slice(start, end);
+});
+
+function reduceItemsPerPage() {
+  if (settings.value.trackersPerPage > minItemsPerPage) {
+    settings.value.trackersPerPage--;
```
MrMarble, line 119: Mutating a pinia store should be avoided, use the `saveSettings` method or pinia `$patch`
Another option could be to implement `increment` and `decrease` methods

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,460 @@
...
+    itemsPerPage.value--;
+  }
+}
+
+function increaseItemsPerPage() {
+  if (settings.value.trackersPerPage < maxItemsPerPage) {
+    settings.value.trackersPerPage++;
```
MrMarble, line 126: Same as in 119

src/components/FlawAffects.vue
```diff
@@ -0,0 +1,1521 @@
...
+  'affects:refresh': [];
+}>();
+
+const { affects, affectsToDelete } = toRefs(props);
+const hasAffects = computed(() => allAffects.value.length > 0);
+const allAffects = computed(() => affectsToDelete.value.concat(affects.value));
+const savedAffects = clone(affects.value) as ZodAffectType[];
+const affectsNotBeingDeleted = computed(
+  () => affects.value.filter((affect) => !affectsToDelete.value.includes(affect))
+);
```
MrMarble, line 46: If we have to filter `affectsToDelete` from `affects` on line 44, does this means that `allAffects` has duplicated affects?

src/components/FlawAffects.vue
```diff
@@ -0,0 +1,1521 @@
...
+  if (settings.value.affectsPerPage > minItemsPerPage) {
+    settings.value.affectsPerPage --;
+    itemsPerPage.value--;
+  }
+}
+
+function increaseItemsPerPage() {
+  if (settings.value.affectsPerPage < maxItemsPerPage) {
+    settings.value.affectsPerPage ++;
+    itemsPerPage.value++;
```
MrMarble, line 484: If this two variables are in sync, could we remove `itemsPerPage` and only use `settings.value.affectsPerPage`?

src/components/FlawAffects.vue
C-Valen: Thanks for pointing out, been checking it and indeed `affects` is already without affects to delete cause in `useFlawAffectsModel` the affect removal method makes an splice on the affects array and adds it to the `affectsToDelete` so that filter was not actually filtering anything. 

Besides of using `affects` directly as `affectsNotBeingDeleted` prop to `FlawTrackers` I sightly changed the `displayedTrackers` computed property to filter there the `sortedAffects` to those not being deleted, so when an affect is set for removal the corresponding tracker is removed from the table.

Just noting that I don't use `affects` directly there cause `sortedAffects` has already filtered affects, so only trackers for the displayed affects are shown in the table.

src/components/FlawAffects.vue
C-Valen: I didn't do that is cause it would imply to use `settings` value in `usePagination` on my initial implementation, which was making it less reusable and abstract. 

Rechecking this I notice that current implementation does not use `itemsPerPage` under `usePagination` and only use on components was the increment/decrement so it was removed and only the settings `perPage` value is now used

C-Valen:
Update v2: Rebased and addressed points from the last reviews.

src/components/__tests__/__fixtures__/sampleFlaws.ts
```diff
@@ -0,0 +1,82 @@
+import { type ZodFlawType } from '@/types/zodFlaw';
+import { sampleFlawComments_1 } from './sampleFlawComments';
+import { sampleFlawAffects_1 } from './sampleFlawAffects';
+import { sampleFlawCvssScores_1 } from './sampleFlawCvssScores';
+
```
superbuggy, line 5: I think having a function is helpful in cases were a sample flaw could be mutated in a test.

MrMarble (approved):

src/composables/usePagination.ts
```diff
@@ -0,0 +1,41 @@
...
+      } else if (currentPage.value >= totalPages.value - 2) {
+        return [1, '..', ...Array.from({ length: 5 }, (_, i) => totalPages.value - 4 + i)];
+      }
+    } else {
+      for (let i = 1; i <= totalPages.value; i++) {
+        result.push(i);
+      }
+    }
+    return result;
+  });
```
superbuggy, line 22:
```suggestion
  const pages = computed(() => {
    const buttonCount = maxPagesToShow - 2;
    if (totalPages.value > maxPagesToShow) {
      if (currentPage.value > 3 && currentPage.value < totalPages.value - 2) {
        return [1, '..', currentPage.value - 1, currentPage.value, currentPage.value + 1, '..', totalPages.value];
      } else if (currentPage.value <= 3) {
        return [...Array.from({ length: buttonCount }, (_, i) => i + 1), '..', totalPages.value];
      } else if (currentPage.value >= totalPages.value - 2) {
        return [1, '..', ...Array.from({ length: buttonCount }, (_, i) => totalPages.value - 4 + i)];
      }
    }
    return Array.from({ length: totalPages.value }, (_, i) => i + 1);
  });
```

src/composables/useTrackers.ts
```diff
@@ -131,6 +132,12 @@ export function useTrackers(flawUuid: string, affects: Ref<ZodAffectType[]>) {
     }
   }
 
+  function toggleTrackerSelections(affectStreams: UpdateStream[]) {
```
superbuggy, line 135: function updateTrackerSelections(affectStreams: UpdateStream[]) {
```suggestion
  function updateTrackerSelections(affectStreams: UpdateStream[]) {
```

src/components/__tests__/__fixtures__/sampleFlawComments.ts
```diff
@@ -0,0 +1,78 @@
+import type { ZodFlawCommentType } from '@/types/zodFlaw';
+
+export function sampleFlawComments_1(): ZodFlawCommentType[] {
```
superbuggy, line 3:
```suggestion
export function sampleFlawComments(): ZodFlawCommentType[] {
```

src/components/__tests__/__fixtures__/sampleFlawTrackers.ts
```diff
@@ -0,0 +1,96 @@
+import type { ZodTrackerType } from '@/types/zodAffect';
+
+export function sampleFlawTrackers_1(affectIds: string[]): ZodTrackerType[] {
```
superbuggy, line 3:
```suggestion
export function sampleFlawTrackers(affectIds: string[]): ZodTrackerType[] {
```

src/components/__tests__/__fixtures__/sampleFlawCvssScores.ts
```diff
@@ -0,0 +1,32 @@
+import type { ZodFlawCVSSType } from '@/types/zodFlaw';
+
+export function sampleFlawCvssScores_1(flawId: string): ZodFlawCVSSType[] {
```
superbuggy, line 3:
```suggestion
export function sampleFlawCvssScores(flawId: string): ZodFlawCVSSType[] {
```

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,454 @@
...
+.affects-trackers {
+  .pagination-controls {
+    display: flex;
+
+    button {
+      height: 2rem;
+      padding-block: 0;
+    }
+
+    .page-btn {
```
superbuggy, line 341:
```suggestion
    .osim-page-button {
```
to distinguish from bootstrap more

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  .trackers-toolbar {
+    display: flex;
+
+    button {
+      height: 2rem;
+      padding-block: 0;
+
+      &.page-btn:disabled {
```
superbuggy, line 360: see abopve

src/components/FlawTrackers.vue
```diff
@@ -0,0 +1,454 @@
...
+        color: $info;
+      }
+    }
+
+    .tracker-badges {
+      display: flex;
+      margin-block: 0;
+      gap: .25rem;
+
+      .per-page-btn {
```
superbuggy, line 371: see above

src/components/FlawAffects.vue
```diff
@@ -0,0 +1,1517 @@
...
+  case displayModes.MODIFIED:
+    return modifiedAffects.value;
+  case displayModes.DELETED:
+    return affectsToDelete.value;
+  case displayModes.CREATED:
+    return newAffects.value;
+  default:
+    return allAffects.value;
+  }
+});
```
superbuggy, line 119:
```suggestion

type DisplayMode = keyof typeof displayModes;
const displayMode = ref<DisplayMode>(displayModes.ALL);
const displayedAffects = computed(() =>  displayModes[displayMode.value] || displayModes.ALL);
```

superbuggy (changes requested):

src/components/__tests__/__fixtures__/sampleFlawComments.ts
C-Valen: Hey @superbuggy I have changed my mind a bit on this point, to a medium preference for permiting an exception on our Camel case notion convention for this sample data cases.

The reason is that removing the underscore makes less clear which number of the sample data it's being used.
For example: `sampleFlawAffects_1` vs `sampleFlawAffects1`, `sampleFlawComments_0` vs `sampleFlawComments0`

We can discuss it in today's code review.

C-Valen:
Update v3: Rebased and addressed points from the last reviews.
@FYI @superbuggy

C-Valen:
Update v3: Fixtures method refactored and test cases adapted accordingly.
FYI @MrMarble @superbuggy

src/composables/usePagination.ts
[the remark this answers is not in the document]
C-Valen: It's functionality is being tested within the affect test suite (see per page test cases), but i think it could be added independent unit test cases on the specific composable implementation.

C-Valen:
Update v4: added test cases for `usePagination` composable.

src/composables/__tests__/usePagination.spec.ts
MrMarble, line 1: Beautiful AAA pattern

MrMarble (approved):

superbuggy (approved):

PR #373, opened 2024-07-31
OSIDB-3214: Save test reports as junit when on CI

MrMarble:
# OSIDB-3214: Save test reports as junit when on CI

In preparation for sending the tests reports via email, I changed the `run_automation_test` script to add parameters to `behave` when running on the pipeline

Closes OSIDB-3214


RichardXuan (approved):
LGTM

PR #375, opened 2024-08-06
Fix the depends on OSIDB API for advance search

huiwangredhat:
Closes OSIDB-3237


features/utils.py
```diff
@@ -51,7 +51,7 @@ def init_remote_firefox_browser():
     profile.set_preference('network.negotiate-auth.trusted-uris', 'https://')
     op = Options()
     op.profile = profile
-    op.add_argument("-headless")
+    #op.add_argument("-headless")
```
MrMarble, line 54: Instead of removing this, we could use the `CI` environment variable so it headless on the pipeline to save on resources
```suggestion
    if 'CI' in os.environ:
        op.add_argument("-headless")
```

MrMarble (approved):

features/utils.py
huiwangredhat: Thanks. Updated.

features/utils.py
```diff
@@ -51,7 +51,8 @@ def init_remote_firefox_browser():
     profile.set_preference('network.negotiate-auth.trusted-uris', 'https://')
     op = Options()
     op.profile = profile
-    op.add_argument("-headless")
+    if 'CI' in os.environ:
```
MrMarble, line 54: also you need to add the `import os` at the top

RichardXuan: os module is already imported, https://github.com/RedHatProductSecurity/osim/pull/375/files#diff-e1542493f59bf0599bc9603c252d7d65e2bb13a000918b15f5a9adbfee4feb97L2

MrMarble: Oh you're right, I didn't saw it

huiwangredhat:
Thanks all.

PR #379, opened 2024-08-09
OSIDB-3170: Splunk integration

MrMarble:
# OSIDB-3170: Splunk integration

## Checklist:

- [x] Commits consolidated
- [ ] Changelog updated
- [x] Jira ticket updated

## Summary:

This pull request implements distributed logging via Splunk.

### Chunks

Since I was modifying `vite.config.ts` I took the opportunity to split the generated bundle in two chunks, one for vendors (node_modules) and other for our own code. This allows us to track the size of our dependencies better as they are now on its own, it also improves caching for the end user, as they don't need to download all the dependencies each time.

### Console

🚀 For now on, on production builds `console.log`, `console.error` and `console.warn` will send the logs to our **splunk** server for helping with identifying user errors, on development builds they will continue to work as expected 

🔒 Splunk needs authentication via token, for that I created a nginx reverse proxy to inject the token, this prevents the token to be exposed to the users and also fixes the cors issues. This uses the environment variable `OSIM_NGINX_PROXY_SPLUNK` for the server url and `OSIM_SPLUNK_TOKEN` for the token itself.

🗑️ I decided to remove `console.debug` from production builds, since now all logs are sent to Splunk, this allows us to have some logs committed to the code that are only visible in development, in case we need them, but on production they are removed completely to keep a clean console output

## Considerations:

We should check all the logs that we are currently using, as some of them aren't that useful, If we want to still keep them, they should be changed to `console.debug` instead. Like this one https://github.com/RedHatProductSecurity/osim/blob/b4ad6850897323a02b5e2a99ecca75824927e618/src/components/widgets/EditableText.vue#L40-L44

I will create a new task for this, to keep this one small

Closes OSIDB-3170


src/shims/console/index.js
```diff
@@ -0,0 +1,59 @@
...
+  fetch(SPLUNK_URL, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+    body: JSON.stringify({
+      event: {
+        message,
+        severity,
+        environment: JSON.parse(localStorage.getItem('UserStore'))?.env || 'unset',
```
MrMarble, line 26: I think we need to add some extra fields so the logs are more useful.
Maybe the user agent (`navigator.userAgent` or a custom string from the `navigator` api) and also some information about the user, like the username to identify the logs, as right now there's no way of telling who did what

MrMarble: Here's a screenshot of the current data 
image

MrMarble:
Also, another good thing about having the logs is that now that we have automated the integration tests, we could see the OSIM logs to know better what happened when one of the tests fails on the pipeline

superbuggy (approved):

C-Valen (approved):
Looks solid, nice work

PR #392, opened 2024-08-29
OSIDB-3373: Reuse access token

MrMarble:
# OSIDB-3373: Reuse access token

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Save access token in memory and use it until it is about to expire.
I added a 1 minute difference to account for clock shift, slow network or other possible problems

## Changes:

* Modified `UserStore` to save the `accessToken`
* Modified `getNextAccessToken` to check the store

## Considerations:

Closes OSIDB-3373

### DIFF
--- a/src/stores/UserStore.ts
+++ b/src/stores/UserStore.ts
@@ -1,9 +1,10 @@
-import { computed, watch } from 'vue';
+import { computed, ref, watch } from 'vue';
 import { defineStore } from 'pinia';
 import { z } from 'zod';
 import jwtDecode from 'jwt-decode';
 import type { JwtPayload } from 'jwt-decode';
 import { useLocalStorage } from '@vueuse/core';
+import { DateTime } from 'luxon';
 
 // const router = useRouter();
 import router from '@/router';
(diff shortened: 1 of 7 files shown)


src/stores/UserStore.ts
```diff
@@ -67,6 +71,23 @@ export const useUserStore = defineStore('UserStore', () => {
...
+    if (!exp) {
+      return true;
+    }
+
+    // get timestamp in seconds minus 1 minute to account for clock drift
+    const now = Math.floor(DateTime.now().minus({ minutes: 1 }).toSeconds());
+    const expiration = Math.floor(DateTime.fromSeconds(exp).toSeconds());
+
+    return now >= expiration;
+  });
```
C-Valen, line 89: Possible property simplification?

```suggestion
const isAccessTokenExpired = computed<boolean>(() => {
const exp = accessToken.value ? jwtDecode<JwtPayload>(accessToken.value)?.exp : null;
return !exp || DateTime.now().toSeconds() >= exp - 60;
});
```

C-Valen (commented):
Looks good. Super minor comments to consider.

src/stores/UserStore.ts
MrMarble: Nice! It was too verbose before.
I also added a try/catch since the `jwtDecode` method will throw if the token is invalid

JakubFrejlach (approved):
LGTM!

PR #396, opened 2024-09-02
Allow using default ordering in flaw list page

dmonzonis:
# OSIDB-3187: Sort results for exact matches at the top

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

The flaw list page allows the user to order the results in descending or ascending order by any of the fields in the header column. By default, it orders the results by created date in descending order (newest flaws first).

This commit adds the option to use the default ordering by OSIDB, by toggling between descending, ascending, and no order when clicking on a field in the header row. This is useful when using the search as OSIDB will return the results ordered based on relevancy. For example, when searching for a CVE, we may want to see the closest matches first, instead of the most recent flaws.

## Changes:

When clicking on a header field, instead of just toggling between descending and ascending order on that field, clicking it a third time will remove the ordering.

Closes OSIDB-3187.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # OSIM Changelog
 
 ## [Unreleased]
+### Added
+* Allow using default ordering in flaw list page (`OSIDB-3187`)
+
 ### Changed
 * Improved performance by reusing access token until is expired (`OSIDB-3373`)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Added
 * Add button to Bugzilla on public and private comments
 * `DEFER` is now a possible affect resolution (`OSIDB-3286`, `OSIDB-3288`)
+* Allow using default ordering in flaw list page (`OSIDB-3187`)
```
MrMarble, line 7: This should be inside the `Unreleased section` update your branch with main, #392 added the section

MrMarble (approved):
LGTM!

@C-Valen This will probably conflict with your changes on the advanced search

PR #397, opened 2024-09-02
✨ Add query mode to advance search

C-Valen:
# OSIDB-3088 Add query mode to advanced search

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Provides support for query filters on advanced search using the new library implemented in OSIDB.

## Changes:

- Adds new optional input for the query filter on the advance search view
- Implement the logic for handling the query parameter on searchs or getting params
- Adapt test cases for advanced search by including new parameter
- Adds a modal for query guide and examples on advanced search 

## Considerations:

- This is the base feature for advanced search redesign, and the first to be merged

Closes OSIDB-3088

### DIFF
--- a/src/components/IssueSearchAdvanced.vue
+++ b/src/components/IssueSearchAdvanced.vue
@@ -2,12 +2,16 @@
 import { computed, ref, watch } from 'vue';
 import { flawImpacts, flawSources, flawIncidentStates } from '@/types/zodFlaw';
 import LabelCheckbox from '@/components/widgets/LabelCheckbox.vue';
+import Modal from '@/components/widgets/Modal.vue';
+import QueryFilterGuide from '@/components/QueryFilterGuide.vue';
+import { useModal } from '@/composables/useModal';
 import { flawFields } from '@/constants/flawFields';
 import { useSearchParams } from '@/composables/useSearchParams';
 import { descriptionRequiredStates } from '@/types/zodFlaw';
 import { affectAffectedness } from '@/types/zodAffect';
 import { sort } from 'ramda';
 const { facets, query, removeFacet, submitAdvancedSearch } = useSearchParams();
+const { isModalOpen, openModal, closeModal } = useModal();
 
 const props = defineProps<{
   isLoading: boolean;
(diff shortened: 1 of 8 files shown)


src/components/IssueSearchAdvanced.vue
```diff
@@ -95,14 +100,37 @@ const optionsFor = (field: string) =>
...
 
 <template>
-  <details :open="shouldShowAdvanced" class="osim-advanced-search-container">
-    <summary class="mb-2" @click="shouldShowAdvanced = true">Advanced Search</summary>
+  <details :open="shouldShowAdvanced" class="osim-advanced-search-container container-fluid">
+    <summary class="mb-1" @click="shouldShowAdvanced = true">Advanced Search</summary>
     <form class="mb-2" @submit.prevent="submitAdvancedSearch">
-      <div v-for="(facet, index) in facets" :key="facet.field" class="input-group mb-2">
+      <div v-if="queryFilterVisible" class="input-group my-1">
+        <div type="button" class="form-control bg-secondary text-white pe-none py-0 d-flex" style="max-width: 13.125%;">
```
MrMarble, line 118: Is the issue with the dynamic width we saw during the demo fixed?
The width of the `Query Filter` label changes with the window

src/components/QueryFilterGuide.vue
```diff
@@ -0,0 +1,352 @@
...
+      <div class="module">
+        <h2 id="fields">OSIDB Fields</h2>
+        <p>
+          In a search query, you should reference the osidb model's fields
+          exactly as they're defined. Search query input has an auto-completion
+          feature that pops up automatically and suggests all available options.
+          If you're not sure what the field name is, then pick one of the options displayed.
+        </p>
+        <p>
+          <strong>TODO: OSIDB Fields Table?</strong>
```
MrMarble, line 74: Remove this?

src/composables/useSearchParams.ts
```diff
@@ -42,9 +43,9 @@ export function useSearchParams() {
     // generate search params based on route query
     const params: Record<string, string> = {};
     const parsedRoute = searchQuery.parse(route);
-    search.value = parsedRoute.query.query || '';
+    query.value = parsedRoute.query.query || '';
```
MrMarble, line 46: replacing the `search` with the `query` was intended?

No the search box on the NavBar will use the djangoql query instead of a normal search

src/components/IssueSearchAdvanced.vue
C-Valen: I fixed in other advance search PR, but probably makes sense to add it here.
(Will do tomorrow)

src/composables/useSearchParams.ts
C-Valen: Yes, as the search facets are handled independently on a loop after this.

MrMarble: But now the search bar on the navbar uses the query instead of the old search, so we lost that functionality?
image

C-Valen: Not sure what you mean, the query is being used for the query input value. Both the individual fields search and the query filter are working to me.

What is `query=testing` supposed to return?

superbuggy (approved):

MrMarble (approved):

PR #394, opened 2024-08-30
ESLint config overhaul

MrMarble:

PR #398, opened 2024-09-02
✨ Support saving query filter on default user search

C-Valen:
# OSIDB-3387 Support saving query filter on default user search

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Add support to save query filter on user's default search store.

## Changes:

- Handles query filter if present when saving default search
- Displays query filter badge if present on active default filters from index view
- Fixes a bug where query filter field was not being set if present in router params
- Adapt test cases

## Considerations:

- Depends on https://github.com/RedHatProductSecurity/osim/pull/397

Closes OSIDB-3387

### DIFF
--- a/src/components/__tests__/FlawSearchView.spec.ts
+++ b/src/components/__tests__/FlawSearchView.spec.ts
@@ -134,7 +134,7 @@ describe('FlawSearchView', () => {
     await flushPromises();
     expect(useRouter().replace).toHaveBeenCalled();
     expect(useRouter().replace.mock.calls[0][0])
-      .toStrictEqual({ query: { query: 'test', search: 'search' } });
+      .toStrictEqual({ query: { query: 'test' } });
   });
 
   it('should call saveFilter on save filter button click', async () => {
@@ -199,7 +199,7 @@ describe('FlawSearchView', () => {
     await flushPromises();
     expect(useRouter().replace).toHaveBeenCalled();
     expect(useRouter().replace.mock.calls[0][0])
-      .toStrictEqual({ query: { query: 'test', requires_cve_description: 'REQUESTED', search: 'search' } });
+      .toStrictEqual({ query: { query: 'search', requires_cve_description: 'REQUESTED' } });
   });
 
   it('should call loadFlaws with major_incident_state on search', async () => {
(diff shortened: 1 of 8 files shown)


src/views/IndexView.vue
```diff
@@ -11,15 +11,16 @@ const { issues, isLoading, isFinalPageFetched, total, loadFlaws, loadMoreFlaws }
...
 );
 
 const isDefaultFilterSelected = ref<boolean>(showFilter.value);
 
 const params = computed(() => {
   const paramsObj = {
     ...tableFilters.value,
-    ...isDefaultFilterSelected.value && searchStore.searchFilters
+    ...isDefaultFilterSelected.value && searchStore.searchFilters,
+    ...isDefaultFilterSelected.value && { query: searchStore.queryFilter }
```
MrMarble, line 23: We need to merge #394 to avoid this kind of thing 😆 
```suggestion
    ...isDefaultFilterSelected.value && { query: searchStore.queryFilter },
```

src/stores/SearchStore.ts
```diff
@@ -4,28 +4,33 @@ import { useLocalStorage } from '@vueuse/core';
 import { computed } from 'vue';
 
 export const SearchSchema = z.object({
-  searchFilters: z.record(z.string())
+  searchFilters: z.record(z.string()),
+  queryFilter: z.string()
 });
 
 export type SearchType = z.infer<typeof SearchSchema>;
```
MrMarble, line 11: This zod object is only used as a type (we don't call `.parse` anywhere), I think it should be changed to a interface and remove zod.

Could you change it?

src/composables/useSearchParams.ts
```diff
@@ -48,7 +48,7 @@ export function useSearchParams() {
     if (parsedRoute.query.search) {
       params.search = parsedRoute.query.search;
     }
-    search.value = parsedRoute.query.query || '';
+    query.value = parsedRoute.query.query || '';
```
MrMarble, line 51: Good catch!

src/components/__tests__/FlawSearchView.spec.ts
MrMarble, line 1: This will probably conflict with my refactor on #400

MrMarble (approved):

superbuggy (approved):

PR #402, opened 2024-09-04
OSIDB-3394: Update OpenAPI schema to include new Flaw sources

MrMarble:
# OSIDB-3394: Update OpenAPI schema to include new Flaw sources

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Updated OpenAPI schema to include new `CVEORG` Flaw source

## Considerations:

Closes OSIDB-3394


src/types/zodFlaw.ts
```diff
@@ -13,7 +13,7 @@ import { zodOsimDateTime, ImpactEnumWithBlank, ZodFlawClassification, ZodAlertSc
 import { ZodAffectSchema, type ZodAffectType } from './zodAffect';
 
 export const RequiresDescriptionEnumWithBlank = { '': '', ...RequiresCveDescriptionEnum } as const;
-export const Source642EnumWithBlank = { '': '', ...Source642Enum } as const;
```
C-Valen, line 16: Shouldn't this be changed from `Source642EnumWithBlank` to `Source521EnumWithBlank`

C-Valen (commented):
Quick question, there is no need to update the dropdowns from FlawForm/ Advanced search?

src/types/zodFlaw.ts
MrMarble: Since the name of the enum changes each time, I wanted to do the minimum changes but I'll update it for consistency

MrMarble: As for the dropdowns, this new source is read only, so the users should not be able to manually create a flaw with it. The Advanced search drop down uses this enum directly, so it is already updated

C-Valen (approved):

src/types/zodFlaw.ts
jobselko: Yeah, that changing enum name is a bug in OSIDB :/

PR #404, opened 2024-09-04
OSIDB-3340: Ruse access token

MrMarble:
# OSIDB-3340: Ruse access token

## Checklist:

- [x] Commits consolidated
- [ ] Changelog updated
- [x] Jira ticket updated

## Summary:

Because the `isAccessTokenExpired` was a `computed` property it wasn't updating properly, always returning `false` which caused unauthenticated issues

## Changes:

Removed the `computed` and return the function directly

Closes OSIDB-3340

### DIFF
--- a/src/services/OsidbAuthService.ts
+++ b/src/services/OsidbAuthService.ts
@@ -126,7 +126,7 @@ export async function getNextAccessToken() {
   const url = `${osimRuntime.value.backends.osidb}/auth/token/refresh`;
   const userStore = useUserStore();
   let response;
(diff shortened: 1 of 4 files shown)


C-Valen (approved):
Just curious, why this wasn't working with computed?

MrMarble:
> Just curious, why this wasn't working with computed?

The `computed` updates when one of its **reactive** dependencies changes, here the only dependency is the `accessToken` but is not updating, the expire date is fixed, what is updating is the `Date.now` value, so because the store is only mounted once, the `isAccessTokenExpired` was computed to `false` and never updated

PR #401, opened 2024-09-03
✨ Multifields emptiness advanced search support

C-Valen:
# OSIDB-3389 Multifields emptiness advanced search support

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Test cases added/updated
- [x] Jira ticket updated

## Summary:

Allow search empty / nonempty on advanced search specific fields (ones supported by OSIDB).

## Changes:

- Removes previous checkbox and code for `cve_description` nonempty
- Adds conditional buttons for empty / nonempty searches on supported fields
- Changes empty param value from empty to word `isempty`
- Adds button to clear any advanced search field
- Adds button to clear query filter
- Updates test cases

## Considerations:

- Empty field searches are now using param value `isempty`
- Non empty field searches remain using `nonempty`
- Depends on https://github.com/RedHatProductSecurity/osim/pull/398

Closes OSIDB-3389

PR #407, opened 2024-09-06
🐛 Affects CVSS CRUD operations

C-Valen:
# OSIDB-3397 Fix Affects CVSS CRUD operations

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Fixes the bug of invalid CVSS field on the new affect section.
Implements the possibility to remove CVSS on affects.

## Changes:

- Fix affect's CVSS field
- Adjust affect table columns to correctly display CVSS vectors
- Adds truncation to flaw fields that have larger length than the table cell
- Adds the CVSS score to the affect CVSS vector when not editing it
- Implements the possibility to remove affect's CVSS

## Demo:
https://github.com/user-attachments/assets/253ab093-cbdd-4b5e-820e-d3fa7175fd3f

## Considerations:

- Possibly impacting on https://github.com/RedHatProductSecurity/osim/pull/401 @superbuggy 
- affect cvss "model" logic was placed within the affect component for simplicity (also considering the component will be refactored), but an independent composable with that logic could be considered in the future.

Closes OSIDB-3397

### DIFF
--- a/src/components/FlawAffects.vue
+++ b/src/components/FlawAffects.vue
@@ -26,6 +26,7 @@ const props = defineProps<{
   embargoed: boolean;
   affects: ZodAffectType[];
   affectsToDelete: ZodAffectType[];
+  affectCvssToDelete: Record<string, string>;
   error: Record<string, any>[] | null;
 }>();
 
@@ -37,7 +38,7 @@ const emit = defineEmits<{
   'affects:refresh': [];
 }>();
 
-const { affects, affectsToDelete } = toRefs(props);
+const { affects, affectsToDelete, affectCvssToDelete } = toRefs(props);
 const hasAffects = computed(() => allAffects.value.length > 0);
 const allAffects = computed(() => affectsToDelete.value.concat(affects.value));
 const savedAffects = clone(affects.value) as ZodAffectType[];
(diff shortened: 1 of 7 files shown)


src/components/FlawAffects.vue
```diff
@@ -424,10 +426,49 @@ function selectAffects(event: Event) {
...
+  const cvssVector = useAffectCvss3Vector(affect).value;
+  if (cvssScore && cvssVector) {
+    return `${affectCvss(affect)?.score || ''} ${useAffectCvss3Vector(affect).value}`;
+  } else {
+    return useAffectCvss3Vector(affect).value;
+  }
+}
+
+function useAffectCvss3Vector(affect: ZodAffectType) {
+  return computed({
```
MrMarble, line 444: This computed is doing nothing, each time the `useAffectCvss3Vector` is called a new instance is returned, and because is called inside a `v-for` each time `paginatedAffects` is updated all the affects will trigger the update

src/components/FlawAffects.vue
```diff
@@ -424,10 +426,49 @@ function selectAffects(event: Event) {
...
+      } else if (newValue !== '') {
+        affect.cvss_scores.push({
+          issuer: 'RH',
+          cvss_version: 'V3',
+          comment: '',
+          score: null,
+          vector: newValue,
+          embargoed: props.embargoed,
+          alerts: [],
+        });
```
MrMarble, line 468: Mutating props is a bad practice, in Vue props are **one way only**, this should be changed to an emit and handled in the parent
[One way data flow](https://vuejs.org/guide/components/props.html#one-way-data-flow)
[Mutating object/array props](https://vuejs.org/guide/components/props.html#mutating-object-array-props)

src/composables/useFlawAffectsModel.ts
```diff
@@ -266,6 +270,34 @@ export function useFlawAffectsModel(flaw: Ref<ZodFlawType>) {
...
+            if (cvssId) {
+              await deleteAffectCvssScore(affect.uuid as string, cvssId);
+              cvssScoresRemovedCount++;
+            }
+          }
+        }
+        affectWithRemovedCvssCount++;
+      } catch (error) {
+        console.error('useFlawAffectsModel::saveAffects() Error removing CVSS score(s):', error);
+      }
```
MrMarble, line 291: Using `await` inside a loop will "block" on each iteration, meaning that if the `deleteAffectCvssScore` takes 1 second to complete and we have 5 cvss, this will take 5 seconds as each request will be made sequentially.

We could use `Promise.allSettled` to improve this
```suggestion
      let affectWithRemovedCvssCount = 0;
      let cvssScoresRemovedCount = 0;
      const cvssScoresToDelete = [];
      try {
        for (const affectId of Object.keys(affectCvssToDelete.value)) {
          const affect = flaw.value.affects.find(affect => affect.uuid === affectId);
          if (affect) {
            const cvssId = affectCvssToDelete.value[affectId];
            if (cvssId) {
               cvssScoresToDelete.push(deleteAffectCvssScore(affect.uuid as string, cvssId));
            }
          }
        }
        cvssScoresRemovedCount = (await Promise.allSettled(cvssScoresToDelete)).filter(({ status }) => status === 'fulfilled').length;
        affectWithRemovedCvssCount++;
      } catch (error) {
        console.error('useFlawAffectsModel::saveAffects() Error removing CVSS score(s):', error);
      }
```

MrMarble: Other than that I think this looks good

superbuggy (approved):
This looks solid!

src/composables/useFlawAffectsModel.ts
C-Valen: Suggestion implemented. Thank you!

PR #420, opened 2024-09-16
🐛 OSIDB-3440: Fix tracker links

C-Valen:
# OSIDB-3440 Fix tracker links

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Hot fix for wrong links on trackers table.

## Changes:

- Fix wrong tracker links by using existing `trackerURL` function 
- Add edge case control for trackers without `external_id`
- Update snaphots
- Add `trackerURL` function to test mocks
- Handle edge case for empty strings on `trackerURL` function
- Add default values to `osimRuntime` ref

## Considerations:

- In live environments external_ids shouldn't be empty but for local instances of OSIDB and for posibility of corrupted data it was considered to add that edge case control on the FE.
- It wasn't really necessary to add default 'dummy' values to `osimRuntime`  ref, but I considered they are useful as example and that they shouldn't be empty strings under any case. Also considering they are used on: 
[Local Configuration](https://github.com/RedHatProductSecurity/osim/blob/main/docs/overview.md#local-configuration)

Closes OSIDB-3440


src/components/FlawTrackers.vue
```diff
@@ -302,9 +303,16 @@ function increaseItemsPerPage() {
...
+                <a
+                  v-if="tracker.external_system_id"
+                  :href="trackerUrl(tracker.type, tracker.external_system_id)"
+                  target="_blank"
+                >
                   {{ `${tracker.external_system_id} ` }}<i class="bi-box-arrow-up-right" />
-                </RouterLink>
+                </a>
+                <span v-else title="This tracker doesn't have External ID">
+                  Undefined
```
MrMarble, line 314: I would prefer to use `--` or just empty when there's no ID, `Undefined` seems like a bug as we don't use it anywhere else

src/services/TrackerService.ts
```diff
@@ -93,9 +93,13 @@ export async function getTrackersForFlaws(requestBody: TrackersFilePost) {
 export function trackerUrl(type: string, id: string): string {
   switch (type) {
     case 'BUGZILLA':
-      return (new URL(`/${id}`, osimRuntime.value.backends.bugzilla)).href;
+      return osimRuntime.value.backends.bugzilla
+        ? (new URL(`/${id}`, osimRuntime.value.backends.bugzilla)).href
+        : 'http://bugzilla-service:8001';
```
MrMarble, line 98: Maybe an _or_ `||` inside the URL function looks better?
```suggestion
      return (new URL(`/${id}`, osimRuntime.value.backends.bugzilla || 'http://bugzilla-service:8001')).href;
```

src/services/TrackerService.ts
```diff
@@ -93,9 +93,13 @@ export async function getTrackersForFlaws(requestBody: TrackersFilePost) {
...
     case 'JIRA':
-      return (new URL(`/browse/${id}`, osimRuntime.value.backends.jiraDisplay)).href;
+      return osimRuntime.value.backends.jiraDisplay
+        ? (new URL(`/browse/${id}`, osimRuntime.value.backends.jiraDisplay)).href
+        : 'http://jira-service:8002';
```
MrMarble, line 102: also here 

```suggestion
      return (new URL(`/browse/${id}`, osimRuntime.value.backends.jiraDisplay || 'http://jira-service:8002')).href;
```

src/stores/osimRuntime.ts
```diff
@@ -44,10 +44,17 @@ const OsimRuntime = z.object({
...
+  backends: {
+    osidb: 'http://osidb-service:8000',
+    osidbAuth: 'kerberos',
+    bugzilla: 'http://bugzilla-service:8001',
+    jira: 'http://jira-service:8002',
+    errata: 'http://errata-service:8003',
+    jiraDisplay: 'http://jira-service:8002',
+  },
+  osimVersion: { rev: 'dev', tag: 'dev', timestamp: '1970-01-01T00:00:00Z' },
+  error: 'error',
```
MrMarble, line 57: why these changes?

src/components/FlawTrackers.vue
C-Valen: Empty was my first option, but when checking the UI it was feeling weird.

I understand that `undefined` feels like a bug to developers, but I think having empty or random characters can be more confusing to the users than 'undefined', cause it provides direct self-explanatory information.

Not strong preference.

src/services/TrackerService.ts
C-Valen: Totally +1

src/stores/osimRuntime.ts
C-Valen: It wasn't really necessary to add default 'dummy' values to osimRuntime ref, but I considered they are useful as example and that they shouldn't be empty strings under any case. Also considering they are used on:
[Local Configuration](https://github.com/RedHatProductSecurity/osim/blob/main/docs/overview.md?rgh-link-date=2024-09-16T12%3A41%3A58Z#local-configuration)

src/components/FlawTrackers.vue
MrMarble: Its just weird, we don't use it anywhere and to me it seems like the link is broken or something happened and the value is "Undefined" instead of not having a link, maybe the word "None" is better if you don't like the `--`, other than that I think it looks good so I'll give my approve

MrMarble (approved):

src/components/FlawTrackers.vue
C-Valen: Let'g go for `None`

PR #425, opened 2024-09-18
🐛 LabelCollapsible readonly error

C-Valen:
# OSIDB-3445 Fix readonly collapsible attribute

## Checklist:

- [x] Commits consolidated
- [-] Changelog updated
- [x] Jira ticket updated

## Summary:

This pr fixes the problems of collapsible labels not working on local OSIM instances, by refactoring the `LabelCollapsible` component to always handle the collapsed state change events on the parent components.

## Changes:

- Refactor `LabelCollapsible` component to use always emits for state changes
- Update uses of `LabelCollapsible` component

## Demo:

https://github.com/user-attachments/assets/6c243f89-00ac-41b9-8f91-1d3d1ee4c671

## Considerations:

- Considered the previously 'dual' way to handle collapse state changes (on parents or local) was aggregating unnecessary complexity, as in all the cases handle the events on the parent is possible.
- The components where the `LabelCollapsible` is used are:
	- FlawAffects (modules filter)
	- FlawAlerts (Alert, AlertList and AlertSection)
	- EditableList (Acknowledgements and References)
	- CvssExplainForm (conditionally displayed within the flaw form)
- No changelog update as the bug is only affecting local OSIM instances

Closes OSIDB-3445

### DIFF
--- a/src/components/CvssExplainForm.vue
+++ b/src/components/CvssExplainForm.vue
@@ -20,7 +20,7 @@ const isExpanded = ref(false);
     :isExpanded="isExpanded"
     label="Explain non-obvious CVSSv3 score metrics"
     class="ms-2 cvss-score-mismatch"
-    @setExpanded="isExpanded = !isExpanded"
+    @toggleExpanded="isExpanded = !isExpanded"
   >
     <textarea
       v-model="modelValue.cvss_scores[rhCvss].comment"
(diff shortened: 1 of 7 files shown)


superbuggy (dismissed):
I think it is probably simplest to have it externally managed, especially with Alerts-focusing behavior.

MrMarble (approved):

PR #426, opened 2024-09-19
OSIDB-3421: Use msw-auto-mock to generate mock data from OpenAPI schema

MrMarble:
# OSIDB-3421: Use msw-auto-mock to generate mock data from OpenAPI schema

## Checklist:

- [x] Commits consolidated
- ~Changelog updated~
- [x] Jira ticket updated

## Summary:

Use msw-auto-mock library to generate mock handlers for [msw](https://mswjs.io/) from the openapi schema.

## Changes:

- The `openapi.yml` file was a symbolic link to `../osidb/openapi.yml` which meant that you needed to have osidb project on that route and also keep it updated for the schema to be correct. I changed this to instead use a simple script to download the latest version from the repository, making sure is always updated. This could be chained together with a Github Workflow that automatically creates a pull request when a new version is out.
- The msw-auto-mock library creates handlers for all the endpoints defined on the schema, I've filtered only **flaw, tracker** and **auth** endpoints, this can be changed through the command in `package.json`. The intention is to have random data with the correct structure to test against. Right now some fields have the wrong type, which should be fixed on OSIDB-3451
- With this, is possible to use mocked data while navigating, this can be useful for integration tests, making them faster and not rely on OSIDB stability, I've decided not to include the code in the pull request as it is not used at the moment, but we just need to add this to `main.ts`
```ts
async function enableMocking() {
  if (import.meta.env.MODE !== 'integration') {
    return;
  }
  const { worker } = await import('./mock-server/browser');

  return worker.start();
}

enableMocking().then(() => app.mount('#app'));
``` 

## Considerations:

Remove `TODO` comments afet OSIDB-3451 is closwd

Closes OSIDB-3421


src/__tests__/setup.ts
```diff
@@ -48,7 +48,7 @@ afterEach(() => {
   expect(onUnhandledRequest).toHaveBeenCalledTimes(0);
 
   // Reset any runtime handlers tests may setup
-  server.resetHandlers();
+  server.restoreHandlers();
```
C-Valen, line 51: May be the comment should be also updated?

C-Valen (approved):

C-Valen (dismissed):

src/services/TrackerService.ts
```diff
@@ -1,9 +1,10 @@
...
 
+import type { TrackerType } from '@/generated-client';
 import { osidbFetch } from '@/services/OsidbAuthService';
 import { osimRuntime } from '@/stores/osimRuntime';
 
-export async function getTracker(uuid: string) {
+export async function getTracker(uuid: string): Promise<TrackerType> {
   return osidbFetch({
     method: 'get',
     url: `/osidb/api/v1/trackers/${uuid}`,
```
superbuggy, line 10: We can probably just axe this function, maybe stick to the ZodTypes for now?

superbuggy (dismissed):

C-Valen (approved):

superbuggy (approved):

PR #452, opened 2024-10-08
OSIDB-3400: Show CVSS comment field when there's a comment

MrMarble:
# OSIDB-3400: Show CVSS comment field whe there's a comment

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

The CVSS comment field explaining the difference between RH and NIST was only shown when both issuer had different scores.  The users complaint that after updating the scores to match, the field was hiding, making it impossible to remove the now unneeded explanation.

## Changes:

Added a condition to the `shouldDisplayEmailNistForm` variable that checks if the comment field is filled and shows it even when both issuer have the same score

## Considerations:

Fixes OSIDB-3400

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Fixed
 * Corrected wrong tooltips on advance search, empty/non-empty buttons (`OSIDB-3502`)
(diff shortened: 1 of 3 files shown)


C-Valen (approved):
Nice, good catch

PR #459, opened 2024-10-16
Add new flaw reference type "Upstream"

dmonzonis:
# OSIDB-3566: Add new flar reference type "Upstream"

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Allow selecting the new "Upstream" flaw reference type that is now
available in OSIDB.

## Changes:

- Add flaw reference type to be chosen from the list.
- Update OpenAPI schema and re-generate client.

Closes OSIDB-3566


C-Valen (commented):
In the `zodFlaw.ts` there is a `FlawREferenceType` that I think should be updated.
Currently is defined as follows:
```
export const FlawReferenceType = {
    Article: 'ARTICLE',
    External: 'EXTERNAL',
    Source: 'SOURCE'
} as const;
```

src/components/IssueFieldReferences.vue
MrMarble, line 1: While reviewing your PR on the code-review call we found out that  the `SOURCE` reference type is not showing on the selector.
image
The proposed change is to add it as **read only**, so the value is shown but the user can't select it, this also adds the `Source` badge (if you have a better name, go for it)
image


---

To apply the change, you need to add the `SOURCE` to https://github.com/RedHatProductSecurity/osim/blob/09e299b39bdd322f088f388e62c63d25abeb6489/src/components/IssueFieldReferences.vue#L30-L35
and update the selectors to use the `LabelSelect` component:

Remove this
https://github.com/RedHatProductSecurity/osim/blob/09e299b39bdd322f088f388e62c63d25abeb6489/src/components/IssueFieldReferences.vue#L115-L124
and this 
https://github.com/RedHatProductSecurity/osim/blob/09e299b39bdd322f088f388e62c63d25abeb6489/src/components/IssueFieldReferences.vue#L115-L124

And use this instead

```vue
<LabelSelect
  v-model="items[itemIndex].type"
  label="Reference type"
  :error="null"
  :options="Object.fromEntries(flawReferenceTypeValues.map(type=> [referenceTypeLabel(type), type]))"
  :options-hidden="excludedReferenceTypes"
/>
```

This is not required and out of the scope of your task so if you don't feel like doing it, I can create a task for it and do it myself, just wanted to give you the option before jumping in with the change.

MrMarble (approved):

C-Valen (dismissed):

superbuggy (approved):

src/components/IssueFieldReferences.vue
dmonzonis: It did seem strange to me that it didn't appear, but looking at the code I saw it was explicitly excluded:
https://github.com/RedHatProductSecurity/osim/blob/09e299b39bdd322f088f388e62c63d25abeb6489/src/components/IssueFieldReferences.vue#L25C1-L28C3

Before changing anything, I would like to know the reason why this was done this way, just to make sure.

MrMarble: Well, this was done way before I joined so I'm not sure, maybe @superbuggy can shed some light on this.

The proposed change will show the `Source` label but, it will not be among the selection options, so users will not be able to select it manually. It is the same functionality but showing the label instead of a blank space.

superbuggy: I think it was never meant to be used in OSIM by the end-user, so we excluded it here

MrMarble (approved):

C-Valen (approved):

PR #461, opened 2024-10-17
OSIDB-3546: CVSS Section extension

C-Valen:
# OSIDB-3546: CVSS Section extension

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Adds new CVSS versions and issuers to the flaw form.

## Changes:

- Introduce new CVSS Section component
- Adds a collapsible section on the flaw form to display other CVSS items if available
- Make the NVD CVSS be displayed always that is available
- Adds coverage files to lint ignore config

## Demo

https://github.com/user-attachments/assets/8097a53e-22da-4c60-95a3-8646b0f0575f

*UPDATE:
- Added new background style to the collapsible section to differentiate it from default flaw form inputs.
- Added `issuerLabels` mapping to `useFlawCvssScores` composable to use custom displaying names (currently only used to change 'NIST' to 'NVD'
- Sightly increased arrow btn size
- Removed border on arrow btn (was being displayed on activation)

image

## Considerations:

- The collapsible section on the form was considered as best option for the CVSS Section as it keeps current layout while adding the required new fields.
- The maximum amount of CVSS Items to be displayed is 12

Closes OSIDB-3546


jobselko:
We should keep issuer names consistent. So either `NVD` or `NIST`, but not both. `NVD` is already in use, so I would probably go with that. Otherwise LGTM!

C-Valen:
> We should keep issuer names consistent. So either `NVD` or `NIST`, but not both. `NVD` is already in use, so I would probably go with that. Otherwise LGTM!

Good point! I agree, we should use consistent naming. Will update that.

superbuggy:
We could possibly visually group the CVSS fields together with a specific background color or some functionally equivalent design concept

MrMarble (dismissed):

superbuggy:
> We should keep issuer names consistent. So either `NVD` or `NIST`, but not both. `NVD` is already in use, so I would probably go with that. Otherwise LGTM!

Will `IssuerEnum` change in osidb?

https://github.com/RedHatProductSecurity/osidb/blob/d2b7a242b19de21fd1ff07b29fd21bdb86478332/osidb/models/abstract.py#L86-L90

C-Valen:
> > We should keep issuer names consistent. So either `NVD` or `NIST`, but not both. `NVD` is already in use, so I would probably go with that. Otherwise LGTM!
> 
> Will `IssuerEnum` change in osidb?
> 
> [RedHatProductSecurity/osidb@`d2b7a24`/osidb/models/abstract.py#L86-L90](https://github.com/RedHatProductSecurity/osidb/blob/d2b7a242b19de21fd1ff07b29fd21bdb86478332/osidb/models/abstract.py#L86-L90)

I don't think so, that's the reason I made a maping to use our own display labels.

src/components/CvssSection.vue
```diff
@@ -0,0 +1,88 @@
...
+  highlightedNvdCvss3String: { char: null | string; isHighlighted: boolean }[][];
+  nistcvss: string;
+  shouldDisplayEmailNistForm: boolean;
+  summary: string;
+}>();
+
+const showAllCvss = ref(false);
+
+const otherCvss = computed(() => props.allCvss.filter(cvssItem =>
+  (!(cvssItem.cvss_version === 'V3' && (cvssItem.issuer === 'RH' || cvssItem.issuer === 'NIST')))));
```
superbuggy, line 25: Since `IssuerEnum` won't change, I wonder if we should use that imported value here `IssuerEnum.NIST`

C-Valen: Yes, I guess we could use the native enum in all the issuers.

src/components/CvssSection.vue
```diff
@@ -0,0 +1,88 @@
...
+</script>
+
+<template>
+  <div>
+    <LabelDiv label="NVD CVSSv3" class="mb-2">
```
MrMarble, line 30: Why do we need a double div here?

MrMarble (approved):

src/composables/useFlawCvssScores.ts
```diff
@@ -5,9 +5,17 @@ import { groupWith, equals } from 'ramda';
...
 
 const formatScore = (score: any) => score?.toFixed(1);
 
+export const issuerLabels: Record<string, string> = {
+  NIST: 'NVD',
+  RH: 'RH',
+  CVEORG: 'CVEOrg',
+  OSV: 'OSV',
+};
+
```
superbuggy, line 18:
```suggestion
export const issuerLabels: Record<string, string> = {
  [IssuerEnum.Nist]: 'NVD',
  [IssuerEnum.Rh]: 'RH',
  [IssuerEnum.Cveorg]: 'CVEOrg',
  [IssuerEnum.Osv]: 'OSV',
};
```

superbuggy (approved):

src/composables/useFlawCvssScores.ts
```diff
@@ -5,9 +5,17 @@ import { groupWith, equals } from 'ramda';
 import { deleteFlawCvssScores, putFlawCvssScores, postFlawCvssScores } from '@/services/FlawService';
 import type { ZodFlawType } from '@/types/zodFlaw';
 import { deepCopyFromRaw } from '@/utils/helpers';
+import { IssuerEnum } from '@/generated-client';
 
 const formatScore = (score: any) => score?.toFixed(1);
 
+export const issuerLabels: Record<string, string> = {
```
MrMarble, line 12: Why are we creating the enum here if it is only used on `CvssSection`?
Wouldn't it make more sense to declare it in that file since it is not used elsewhere?

C-Valen: IMO it is well placed, as the composable is specific for CVSS and we keep the component script logic minimized.
No strong preference tho.

MrMarble (approved):

MrMarble (dismissed):

src/components/CvssSection.vue
```diff
@@ -0,0 +1,89 @@
...
+
+import { issuerLabels } from '@/composables/useFlawCvssScores';
+
+import type { ZodFlawCVSSType } from '@/types/zodFlaw';
+import { IssuerEnum } from '@/generated-client';
+
+const props = defineProps<{
+  allCvss: ZodFlawCVSSType[];
+  bugzilla: string;
+  cveid: null | string;
```
superbuggy:
```suggestion
  cveId: null | string;
```

src/components/FlawForm.vue
```diff
@@ -274,31 +273,16 @@ const createdDate = computed(() => {
...
-                      />
-                    </div>
-                  </div>
-                </LabelDiv>
-              </div>
-            </div>
+            <CvssSection
+              :highlightedNvdCvss3String
+              :shouldDisplayEmailNistForm
+              :cveid="flaw.cve_id"
```
superbuggy:
```suggestion
              :cveId="flaw.cve_id"
```

src/components/FlawForm.vue
```diff
@@ -274,31 +273,16 @@ const createdDate = computed(() => {
...
+              :summary="flaw.comment_zero"
+              :bugzilla="bugzillaLink"
+              :cvss="rhCvss3String"
+              :allCvss="flaw.cvss_scores"
+              :nistcvss="nvdCvss3String"
```
superbuggy:
```suggestion
              :nistCvss="nvdCvss3String"
```

src/components/CvssSection.vue
```diff
@@ -0,0 +1,89 @@
...
+            <template v-for="(chars, index) in highlightedNvdCvss3String" :key="index">
+              <span v-if="chars[0].isHighlighted" class="text-primary">
+                {{ chars.map(c => c.char).join('') }}
+              </span>
+              <template v-else>{{ chars.map(c => c.char).join('') }}</template>
+            </template>
+          </template>
+          <template v-else>
+            <span>
+              {{ nistcvss }}
```
superbuggy:
```suggestion
              {{ nistCvss }}
```

src/components/CvssSection.vue
```diff
@@ -0,0 +1,89 @@
...
+            </span>
+          </template>
+        </div>
+        <div v-if="shouldDisplayEmailNistForm" class="col-auto align-self-center">
+          <CvssNISTForm
+            :cveid
```
superbuggy:
```suggestion
            :cveId
```

superbuggy (changes requested):
Some non-camelCase names could be renamed

superbuggy (approved):

PR #462, opened 2024-10-17
History section UI update

C-Valen:
# History section UI update

## Checklist:

- [x] Commits consolidated
- [-] Changelog updated
- [-] Jira ticket updated

## Summary:

PR to address small feedback changes to the history section UI

## Changes:

- Moves the history section at the end of the flaw
- Increases section title font size

## Considerations:

- This PR will be merge into UAT and kept as draft until receiving and addressing more feedback from stakeholders.

### DIFF
--- a/src/components/FlawForm.vue
+++ b/src/components/FlawForm.vue
@@ -223,7 +223,6 @@ const createdDate = computed(() => {
               Open in Bugzilla <i class="bi-box-arrow-up-right ms-2" />
             </a>
           </div>
-          <FlawHistory :flaw="flaw" @expandFocusedComponent="expandFocusedComponent" />
           <FlawAlertsList
             :flaw="flaw"
             class="col-12 osim-alerts-banner"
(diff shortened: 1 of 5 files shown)


src/components/__tests__/__snapshots__/FlawForm.spec.ts.snap
```diff
@@ -839,6 +838,7 @@ This module has no trackers associated"><i data-v-0fd8a3c0="" class="bi bi-excla
       </div>
     </section>
   </div>
+  <!--v-if-->
```
superbuggy, line 841:
```suggestion
```

src/components/FlawForm.vue
```diff
@@ -466,6 +465,7 @@ const createdDate = computed(() => {
             @loadInternalComments="loadInternalComments"
           />
         </div>
+        <FlawHistory :flaw="flaw" @expandFocusedComponent="expandFocusedComponent" />
```
MrMarble, line 468: Maybe we should add a check like we do with the comments so this section is hidden when creating a flaw?
```suggestion
        <FlawHistory v-if="mode === 'edit'" :flaw="flaw" @expandFocusedComponent="expandFocusedComponent" />
```

superbuggy (approved):

src/components/FlawForm.vue
MrMarble: After further discussion, we decided to leave as it

C-Valen: Off thread discussion output:
Makes sense to use conditional render checks inside the component when the check data is available.
In the comments section the data for the render check comes from a previous render request, that's the reason the render condition is specified in the parent (edit mode).

PR #468, opened 2024-10-28
OSIDB-3596: Allow empty impact of flaws

MrMarble:
# OSIDB-3596: Allow empty impact of flaws

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Changed flaw validation to allow empty Impact. This was already supported by OSIDB but not on OSIM

## Changes:

*  We were only using the values of the `ImpactEnum`, I've change it so now we use the keys as label
* Added a `withBlank` prop to the `labelSelect` component to show an empty "default" value. I did this so we don't need to keep track of this kind of values on enums, as they are invalid, just needed for UX

## Considerations:

Closes OSIDB-3596


src/types/zodFlaw.ts
```diff
@@ -194,7 +200,7 @@ export const ZodFlawSchema = z.object({
   }
 
   if (zodFlaw.classification?.state !== 'REJECTED') {
-    if (!zodFlaw.impact) {
+    if (zodFlaw.impact === null) {
```
superbuggy:
```suggestion
    if (zodFlaw.impact === null || zodFlaw.impact === undefined ) {
```

superbuggy (approved):

PR #469, opened 2024-10-29
add case for advance search: Search flaws with both query filter and …

RichardXuan:
…selected field. And some little fix for failed cases.

Fixes OSIDB-3540


features/steps/advance_search.py
```diff
@@ -218,3 +218,15 @@ def step_impl(context):
...
+
+
+@then('I am able to search flaws with both query filter and selected field')
+def step_impl(context):
+    cve_id = get_data_from_tmp_data_file(FLAW_ID_KEY)
+    advanced_search_page = AdvancedSearchPage(context.browser)
+    # set query in query filter
+    advanced_search_page.set_query_filter(f'cve_id = "{cve_id}"')
+    # set query in selected field
+    advanced_search_page.select_field_and_value_to_search("cve_id", cve_id)
```
MrMarble, line 230: Should we use different fields for the query and the form?

RichardXuan: Make sense. I will change this. Thanks.

RichardXuan: Updated.

MrMarble (approved):

PR #472, opened 2024-11-01
🐛 Refresh affect trackers after filing trackers

superbuggy:
# OSIDB-3483 Fix affect tracker refreshing after filing trackers

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

[Replace with a brief summary of the changes introduced by this PR.]

## Changes:

## Considerations:

Closes OSIDB-3483

### DIFF
--- a/src/components/FlawAffects/FlawAffects.vue
+++ b/src/components/FlawAffects/FlawAffects.vue
@@ -534,7 +534,6 @@ const displayedTrackers = computed(() => {
             :relatedFlaws="relatedFlaws"
             :flaw="flaw"
             :specificAffectsToTrack="specificAffectsToTrack"
-            @affects-trackers:refresh="refreshAffects"
           />
         </template>
       </Modal>
--- a/src/components/FlawAffects/FlawAffects.vue
+++ b/src/components/FlawAffects/FlawAffects.vue
@@ -30,7 +30,7 @@ const props = defineProps<{
   relatedFlaws: ZodFlawType[];
 }>();
 
-const flaw = useFlaw();
+const { flaw } = useFlaw();
 
 const affectsEditingStore = useAffectsEditingStore();
 const {
--- a/src/composables/useFetchFlaw.ts
+++ b/src/composables/useFetchFlaw.ts
@@ -1,23 +1,17 @@
 import { ref } from 'vue';
 
+import { useFlaw } from '@/composables/useFlaw';
 import { resetInitialAffects } from '@/composables/useFlawAffectsModel';
 
 import { getFlaw, getRelatedFlaws } from '@/services/FlawService';
-import type { ZodAffectType, ZodFlawType } from '@/types';
+import type { ZodAffectType } from '@/types';
 import { useToastStore } from '@/stores/ToastStore';
 import { getDisplayedOsidbError } from '@/services/osidb-errors-helpers';
 
-const flaw = ref<null | ZodFlawType>(null);
-const relatedFlaws = ref<ZodFlawType[]>([]);
 const isFetchingRelatedFlaws = ref(false);
 
-export function initializeFlaw() {
-  flaw.value = null;
-  relatedFlaws.value = [];
-  resetInitialAffects();
-}
-
 export function useFetchFlaw() {
+  const { flaw, relatedFlaws, resetFlaw } = useFlaw();
   const { addToast } = useToastStore();
 
   const didFetchFail = ref<boolean>(false);
@@ -45,7 +39,7 @@ export function useFetchFlaw() {
     } catch (error) {
       console.error('useFetchFlaw::fetchFlaw()', error);
       didFetchFail.value = true;
-      flaw.value = null;
+      resetFlaw();
       resetInitialAffects();
       addToast({
         title: 'Error loading Flaw',
--- a/src/composables/useFlaw.ts
+++ b/src/composables/useFlaw.ts
@@ -11,10 +11,10 @@ function resetFlaw() {
 }
 
 export function useFlaw() {
-  return { flaw, relatedFlaws, blankFlaw, resetFlaw };
+  return { flaw, relatedFlaws, resetFlaw };
 }
 
-function blankFlaw(): ZodFlawType {
+export function blankFlaw(): ZodFlawType {
   return {
     affects: [],
     classification: {
--- a/src/composables/useFlaw.ts
+++ b/src/composables/useFlaw.ts
@@ -3,12 +3,18 @@ import { ref } from 'vue';
 import type { ZodFlawType } from '@/types/zodFlaw';
 
 const flaw = ref<ZodFlawType>(blankFlaw());
+const relatedFlaws = ref<ZodFlawType[]>([]);
 
-export function useFlaw() {
-  return flaw;
+function resetFlaw() {
+  flaw.value = blankFlaw();
+  relatedFlaws.value = [];
 }
 
-export function blankFlaw(): ZodFlawType {
+export function useFlaw() {
+  return { flaw, relatedFlaws, blankFlaw, resetFlaw };
+}
+
+function blankFlaw(): ZodFlawType {
   return {
     affects: [],
     classification: {
--- a/src/composables/useFlawModel.ts
+++ b/src/composables/useFlawModel.ts
@@ -32,7 +32,7 @@ import {
 import { createSuccessHandler, createCatchHandler } from './service-helpers';
 
 export function useFlawModel(forFlaw: ZodFlawType, onSaveSuccess: () => void) {
-  const flaw = useFlaw();
+  const { flaw } = useFlaw();
   flaw.value = forFlaw;
 
   const isSaving = ref(false);
--- a/src/composables/useRelatedFlawTrackers.ts
+++ b/src/composables/useRelatedFlawTrackers.ts
@@ -258,6 +258,8 @@ export function useRelatedFlawTrackers(
   async function refreshRelatedFlaws() {
     isLoadingTrackers.value = true;
     // After filing new trackers, there is a delay before the new trackers are available on the flaws
+    // This is a temporary workaround that may not work in all situations
+    // TODO: Implement a sustainable solution (OSIDB-3961)
     await new Promise(resolve => setTimeout(resolve, 2000));
     try {
       for (const flaw of selectedRelatedFlaws.value) {
--- a/src/composables/useRelatedFlawTrackers.ts
+++ b/src/composables/useRelatedFlawTrackers.ts
@@ -257,6 +257,8 @@ export function useRelatedFlawTrackers(
 
   async function refreshRelatedFlaws() {
     isLoadingTrackers.value = true;
+    // After filing new trackers, there is a delay before the new trackers are available on the flaws
+    await new Promise(resolve => setTimeout(resolve, 2000));
     try {
       for (const flaw of selectedRelatedFlaws.value) {
         const fetchedFlaw = await getFlaw(flaw.uuid);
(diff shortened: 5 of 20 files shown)


src/composables/useFlawModel.ts
```diff
@@ -239,3 +240,15 @@ function flawErrors(flaw: ZodFlawType) {
 
   return mirroredFlaw;
 }
+
+export async function refreshFlawAffectTrackers(flaw: Ref<ZodFlawType>) {
+  const freshFlaw = await getFlaw(flaw.value.uuid);
```
MrMarble, line 245: Should we add some kind of loader to this action?

superbuggy: Hrm, maybe we should the delay should be long just to grab the flaw, but that might be best to trigger a loading state

src/composables/useFlaw.ts
MrMarble, line 1: Since this is now working like a store, could we rewrite it so it feels more like it? It will also follow the convention we use for composables export a function that returns an object, this also incorporates `useRelatedFlaws` which as the name suggest is related, an as I see in your PR you use both at the same time:

```typescript
const flaw = ref<null | ZodFlawType>(blankFlaw());
const relatedFlaws = ref<ZodFlawType[]>([]);

// I think I'd prefer to use 'resetFlaw' as initialization is what `blankFlaw` does
function initializeFlaw { 
  flaw.value = null;
  relatedFlaws.value = [];
}

export function useFlaw() {
  return { flaw, relatedFlaws, initializeFlaw }
}

```

src/composables/useFetchFlaw.ts
```diff
@@ -1,21 +1,17 @@
...
-import type { ZodAffectType, ZodFlawType } from '@/types';
+import type { ZodAffectType } from '@/types';
 import { useToastStore } from '@/stores/ToastStore';
 import { getDisplayedOsidbError } from '@/services/osidb-errors-helpers';
 
-const flaw = ref<null | ZodFlawType>(null);
-const relatedFlaws = ref<ZodFlawType[]>([]);
-const isFetchingRelatedFlaws = ref(false);
+const flaw = useFlaw();
+const relatedFlaws = useRelatedFlaws();
```
MrMarble, line 12: As an example with the proposed change, this could be:
```suggestion
const { flaw, relatedFlaws, initializeFlaw } = useFlaw();
```
> Remove the `initializeFlaw` import at the top

MrMarble (changes requested):
I don't think we should enforce the `const flaw = useFlaw() as Ref<ZodFlawType>;` everywhere.
The only place where `initializeFlaw` is called is on fetchFlaw, maybe instead of using the global flaw, use a local variable and only if the flaw is fetched correctly, update the global one, that way we can remove the optional null on `useFlaw` and leave it like `ref<ZodFlawType>(blankFlaw());` so we can remove all the type casting.

src/components/FlawAffects/FlawAffects.vue
```diff
@@ -30,7 +30,7 @@ const props = defineProps<{
   relatedFlaws: ZodFlawType[];
 }>();
 
-const flaw = useFlaw();
+const flaw = useFlaw() as Ref<ZodFlawType>;
```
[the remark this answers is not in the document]
superbuggy: It's also called when `FlawEditView` unmounts,  https://github.com/RedHatProductSecurity/osim/blob/5bc3463bb186ee7c9f9619e7af3248c51490898d/src/views/FlawEditView.vue#L16

but I think your suggestion probably works. The `null` was there to avoid having a blank flaw potentially fed into the Editing view, which seems like a more natural convention. I don't like the assertion-based approach I used, but it seemed like the more natural use of an assertion until, in the course of revisiting the logic, I had to hoist that assertion to the store/global level.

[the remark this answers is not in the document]
MrMarble: >It's also called when FlawEditView unmounts

Oh good catch I missed that

> The `null` was there to avoid having a blank flaw potentially fed into the Editing view, which seems like a more natural convention.

The `flawModel` overrides whatever the `useFlaw` composable returns
https://github.com/RedHatProductSecurity/osim/blob/e23c1424a3085f36cf6c69fbde5b334fd043733e/src/composables/useFlawModel.ts#L35-L36
> side note, maybe accept an optional parameter to `useFlaw` for setting the initial value?


My point of view is, if we are certain enough to override the `null` type everywhere (which could fail in runtime since the value might by null and thus property access will fail), why not remove the optional `null` and have it always return the value if we check that **when** is called is always with an expected value.

C-Valen (commented):
I'm getting a type error when trying to file a bunch of trackers:

image

src/composables/useFlaw.ts
```diff
@@ -3,12 +3,18 @@ import { ref } from 'vue';
...
+const relatedFlaws = ref<ZodFlawType[]>([]);
+
+function resetFlaw() {
+  flaw.value = blankFlaw();
+  relatedFlaws.value = [];
+}
 
 export function useFlaw() {
-  return flaw;
+  return { flaw, relatedFlaws, blankFlaw, resetFlaw };
```
MrMarble, line 14: I think we could leave `blankFlaw` outside of the `useFlaw` composable as it is returning a constant value that does not modifies `flaw` in any way

src/composables/useRelatedFlawTrackers.ts
```diff
@@ -142,27 +143,41 @@ export function useRelatedFlawTrackers(
...
-  // Logic for isLoadingTrackers could be moved to useSingleFlawTrackers as an improvement
   watch(affectsBySelectedFlawId, (newRelatedAffects: Record<string, ZodAffectType[]>) => {
+    updateMultiFlawTrackers(newRelatedAffects, false);
+  }, { immediate: true });
+
+  watch(relatedFlaws, () => {
+    selectedRelatedFlaws.value.forEach((selectedFlaw) => {
+      const index = relatedFlaws.value.findIndex(({ uuid }) => uuid === selectedFlaw.uuid);
+      if (index !== -1) {
+        selectedRelatedFlaws.value[index].affects = relatedFlaws.value[index].affects;
```
MrMarble, line 154: This `index` comes from the `relatedFlaws` array, that holds all the related flaws, `selectedRelatedFlaws` is only populated by the ones selected so it will not always match

src/composables/useFlaw.ts
superbuggy: Ah these are reversed!

src/composables/useRelatedFlawTrackers.ts
superbuggy: Ah these are reversed!

src/composables/useFlaw.ts
superbuggy: I had this same thought

src/composables/useRelatedFlawTrackers.ts
```diff
@@ -142,27 +143,41 @@ export function useRelatedFlawTrackers(
...
 
-  // Logic for isLoadingTrackers could be moved to useSingleFlawTrackers as an improvement
   watch(affectsBySelectedFlawId, (newRelatedAffects: Record<string, ZodAffectType[]>) => {
+    updateMultiFlawTrackers(newRelatedAffects, false);
+  }, { immediate: true });
+
+  watch(relatedFlaws, () => {
+    selectedRelatedFlaws.value.forEach((selectedFlaw) => {
+      const index = relatedFlaws.value.findIndex(({ uuid }) => uuid === selectedFlaw.uuid);
+      if (index !== -1) {
```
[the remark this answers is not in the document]
superbuggy: That's cool!

src/composables/useRelatedFlawTrackers.ts
```diff
@@ -142,27 +143,41 @@ export function useRelatedFlawTrackers(
...
 
-  // Logic for isLoadingTrackers could be moved to useSingleFlawTrackers as an improvement
   watch(affectsBySelectedFlawId, (newRelatedAffects: Record<string, ZodAffectType[]>) => {
+    updateMultiFlawTrackers(newRelatedAffects, false);
+  }, { immediate: true });
+
+  watch(relatedFlaws, () => {
+    selectedRelatedFlaws.value.forEach((selectedFlaw, index) => {
+      const relatedIndex = relatedFlaws.value.findIndex(({ uuid }) => uuid === selectedFlaw.uuid);
+      if (index !== -1) {
```
MrMarble, line 153: I think this was left behind after the renaming, `index` comes from the loop so it will never be `-1`
```suggestion
      if (relatedIndex !== -1) {
```

MrMarble: Once this is fixed I'll leave my approval, rest LGTM

C-Valen (approved):

MrMarble (approved):

src/composables/useRelatedFlawTrackers.ts
```diff
@@ -238,10 +255,28 @@ export function useRelatedFlawTrackers(
     }
   }
 
+  async function refreshRelatedFlaws() {
+    isLoadingTrackers.value = true;
+    // After filing new trackers, there is a delay before the new trackers are available on the flaws
+    await new Promise(resolve => setTimeout(resolve, 2000));
```
MrMarble, line 261: I don't think a random timeout is a solution, tracker creation is asynchronous, it depends not only on Jira API response time, but also if workers are idle and ready to take the creation task or busy with periodic work / requests from other users, this value might work locally, but it will surely fail in production.

Maybe we could implement a polling system, make a request every second or half a second to the `/affects/<affect_uuid>` endpoint to see if the tracker is already there, if it takes more than 10 requests (5 to 10 seconds depending on the delay) we show an alert with `Tracker creation took too long` so we don't keep the user waiting. 

What do you think?

MrMarble: Another thing I saw while testing this is that when we make the POST to `trackers` for creation, the response already has everything we need to show the tracker in the UI, can't we use this to update the trackers table? (`status` is empty but since we are creating it, I think we can assume it is `NEW`)
image

superbuggy: I noted this as well, but it does not have the relational associations. Additionally, matching based on `ps_update_stream` will not yield correct associations in some situations.

superbuggy: > I don't think a random timeout is a solution, tracker creation is asynchronous, it depends not only on Jira API response time, but also if workers are idle and ready to take the creation task or busy with periodic work / requests from other users, this value might work locally, but it will surely fail in production.
> 
> Maybe we could implement a polling system, make a request every second or half a second to the `/affects/<affect_uuid>` endpoint to see if the tracker is already there, if it takes more than 10 requests (5 to 10 seconds depending on the delay) we show an alert with `Tracker creation took too long` so we don't keep the user waiting.
> 
> What do you think?

I agree and had considered the same, but have concerns about implementation complexity and resource allocation for this particular PR.

While not ideal, I think this is candidate solution for a temporary fix, though I think providing a warning that a hard refresh may be required. I agree with the polling approach, and while not a massive extra lift, does expand the scope of the solution. I considered this strategy, but we are already experiencing scope creep here. I agree with you, but from an operational perspective, I am somewhat hesitant to implement a more robust solution at this time.

It is worth noting that the JIRA sync is non-blocking for the relational associations to take place and that the 2 seconds seems sufficient for the relational associations to propagate in my testing.

MrMarble (dismissed):

src/composables/useRelatedFlawTrackers.ts
superbuggy: Alright, so pending the discussion between @MrMarble and I during our code review, I realized we can in fact use the data from the request-response cycle for trackers _since each tracker is created one by one_. This however does expand the scope of the PR, and we have fix planned documented in `OSIDB-3961`.

PR #473, opened 2024-11-04
🐛 OSIDB-3645: Use 24h format on history dates

C-Valen:
# OSIDB-3645: Use 24h format on history dates

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Small fix to correct the format of dates in the history section, which was in 12h.
As in all other dates we use 24h time format this will correct the parse function being used by the history section dates.

## Changes:

- Correct time format on `formatDateWithTimezone` helper function
- Adjust history test time to verify format is in 24h

## Demo:
image

Closes OSIDB-3645

### DIFF
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -73,7 +73,7 @@ export function formatDate(date: Date | string, includeTime: boolean): string {
 
 export function formatDateWithTimezone(value: string) {
   return DateTime.fromISO(value, { setZone: true })
-    .toFormat('yyyy-MM-dd hh:mm ZZZZ');
+    .toFormat('yyyy-MM-dd HH:mm ZZZZ');
 }
 
 export function getSpecficCvssScore(scores: any[], issuer: string, version: string) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # OSIM Changelog
 
+## [Unreleased]
+### Fixed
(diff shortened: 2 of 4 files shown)


src/utils/helpers.ts
```diff
@@ -73,7 +73,7 @@ export function formatDate(date: Date | string, includeTime: boolean): string {
 
 export function formatDateWithTimezone(value: string) {
   return DateTime.fromISO(value, { setZone: true })
-    .toFormat('yyyy-MM-dd hh:mm ZZZZ');
+    .toFormat('yyyy-MM-dd HH:MM ZZZZ');
```
MrMarble, line 76: `MM` is for **Months**.
```suggestion
    .toFormat('yyyy-MM-dd HH:mm ZZZZ');
```
https://moment.github.io/luxon/#/formatting?id=table-of-tokens

Notice how the snapshot has changed from `06` (minutes) to `10` (month 10)

MrMarble:
Could you change the hour in the test so the change is more clear?
For example  `15`, so the output is `15` instead of `03` and the change between 12h and 24h is visually detected
 https://github.com/RedHatProductSecurity/osim/blob/a6fe8f9984ba5676662eced8344f620599fc6cf9/src/components/__tests__/FlawHistory.spec.ts#L11

MrMarble (dismissed):

MrMarble (approved):

superbuggy (approved):

PR #476, opened 2024-11-11
OSIDB-3275: Prevent flaw saving if CVSS vector is invalid

MrMarble:
# OSIDB-3275: Prevent flaw saving if CVSS vector is invalid

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

* Added CVSS validation to `validate` method in `useFlawModel` composable to prevent saving the flaw if the vector is invalid
* Renamed `useFlaws.spec.ts` to `useFlawsFetching.spec.ts`
* Moved fixtures to the top level folder for using them outside of the `components` folder
* Added tests to `useFlawModel`

## Considerations:

Closes OSIDB-3275


src/__tests__/__fixtures__/sampleFlawRequired.json
```diff
@@ -1,14 +1,14 @@
...
+    "cve_id": "CVE-2024-0001",
     "impact": "MODERATE",
     "components": [
         "kernel"
     ],
     "title": "Sample title",
     "trackers": [],
     "comment_zero": "Comment Zero == Patient Zero",
-    "cve_description": "",
+    "cve_description": "Some description",
```
C-Valen, line 11: Why was this added?

C-Valen (commented):
Generally looks good. Weak preference to add small scope description to branches.

src/__tests__/__fixtures__/sampleFlawRequired.json
MrMarble: the `requires_cve_description` field has the value `APPROVED`, and that only can happen if the description is not empty, another solution could be to change the value of the `requires_cve_description`

C-Valen: Oh good catch. I'd prefer that second approach, removing the requires state, just cause `cve_description` is not a required field.

superbuggy: If the real-world workflow is that a CVE scoring description is required, then maybe this good to add some example description

superbuggy (dismissed):

src/composables/__tests__/useFlawModel.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+      const { flawRhCvss3, updateFlaw } = mountFlawModel(sampleFlawFull as ZodFlawType);
+
+      flawRhCvss3.value.vector = 'invalid';
+      await flushPromises();
+      updateFlaw();
+
+      expect(putFlaw).not.toHaveBeenCalled();
+    });
+  });
+});
```
superbuggy, line 124: A welcome addition to our coverahe

C-Valen (approved):

PR #478, opened 2024-11-11
🐛♻️ Fix Save all behavior and Reorganize Flaw & Affect App-State

superbuggy:
# OSIDB-3664 Fix Save All Affect Section
# OSIDB-3666 Refactor app state organization
# OSIDB-3700 Fix undesirable sorting of edited affects

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Test cases added/updated
- [x] Jira ticket updated

## Summary:

Fixes the save all button behavior with an extensive refactoring of 

## Changes:

- Fix save all button reverting affects to their original state.
- Fixes an unreported CVSS change tracking bug.
- Merges useAffectSelections and affect editing functions into a single store.
  - Editing functions were moved out of `FlawAffects` and into their own dedicated store module (`AffectsEditingStore`).
  - Selection logic combined with store logic to avoid modules with circular dependencies
    - Uses separate modules internal to the store to maintain organization
    - Uses dependency injection to achieve loosest coupling
  - Separate Flaw data and Flaw Model logic so that dependent stores and composables can use the flaw data without importing unused Flaw Model functionalities.

## Considerations:

The small bug necessitated a fix that presented an opportunity to either accumulate more technical debt, or to pay it off. Accumulated more technical debt would have increased the number of sources of truth for affects data, and posed a significant risk of data inconsistency in the application. This posed an added risk of dramatically increasing the cost of future development work around affects-related features.

Closes OSIDB-3664. Closes OSIDB-3666. Closes OSIDB-3700.

PR #511, opened 2024-11-28
🐛 OSIDB-3708: Fix affect resolution on not affected changes

C-Valen:
# OSIDB-3708: Fix affect resolution on not affected changes

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Affect resolution needs to be set to empty value when the affectedness is set to `NOTAFFECTED`. The bug was preventing this change to be automatic when changing an affect to `NOTAFFECTED`, forcing the users to manually set the resolution to empty prior to changing the affectedness.

## Changes:

- Move the `onChange` event designated to handle the automatic resolution change, from the affectedness option element to the the select element
- Add new test case to cover the bus scenario

## Considerations:

- Depends on https://github.com/RedHatProductSecurity/osim/pull/478

## Demo:

https://github.com/user-attachments/assets/b8c0d0aa-d035-4d35-9b4d-a3cff7e6afda

Closes OSIDB-3708

### DIFF
--- a/src/components/FlawAffects/FlawAffectsTableRow.vue
+++ b/src/components/FlawAffects/FlawAffectsTableRow.vue
@@ -188,13 +188,13 @@ function affectednessChange(event: Event, affect: ZodAffectType) {
         v-model="affect.affectedness"
         class="form-select"
         @keydown="handleKeystroke($event, affect)"
+        @change="affectednessChange($event, affect)"
       >
         <option
           v-for="affectedness in affectAffectedness"
           :key="affectedness"
           :value="affectedness"
           :selected="affectedness === affect.affectedness"
-          @change="affectednessChange($event, affect)"
         >
           {{ affectedness }}
         </option>
--- a/src/components/FlawAffects/__tests__/FlawAffects.spec.ts
+++ b/src/components/FlawAffects/__tests__/FlawAffects.spec.ts
@@ -233,6 +233,30 @@ describe('flawAffects', () => {
     expect(resolutionOptions.includes('DEFER')).toBe(false);
   });
 
+  osimFullFlawTest('Changing affect to not affected automatically sets empty resolution', async () => {
+    const affectsTableRows = subject.findAll('.affects-management table tbody tr');
+
+    let resolutionSelect = affectsTableRows[0].find('td:nth-of-type(6) span');
+    expect(resolutionSelect.text()).not.toEqual('');
+
+    const affectRowEditBtn = affectsTableRows[0].find('td:last-of-type button:first-of-type');
+    expect(affectRowEditBtn.exists()).toBe(true);
+    await affectRowEditBtn.trigger('click');
+
+    const affectsTableEditingRows = subject.findAll('.affects-management table tbody tr.editing');
+    const affectednessSelect = affectsTableEditingRows[0].find('td:nth-of-type(5) select');
+    await affectednessSelect.setValue('NOTAFFECTED');
+    const selectedAffectedness = affectednessSelect.find('option[selected]');
+    expect(selectedAffectedness.text()).toBe('NOTAFFECTED');
+
+    const affectRowCommitBtn = affectsTableEditingRows[0].find('td:last-of-type button:first-of-type');
+    expect(affectRowCommitBtn.exists()).toBe(true);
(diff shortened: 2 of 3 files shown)


src/components/FlawAffects/__tests__/FlawAffects.spec.ts
```diff
@@ -233,6 +233,26 @@ describe('flawAffects', () => {
...
+    await affectednessSelect.setValue('NOTAFFECTED');
+    const selectedAffectedness = affectednessSelect.find('option[selected]');
+    expect(selectedAffectedness.text()).toBe('NOTAFFECTED');
+
+    const affectRowCommitBtn = affectsTableEditingRows[0].find('td:last-of-type button:first-of-type');
+    expect(affectRowCommitBtn.exists()).toBe(true);
+    await affectRowCommitBtn.trigger('click');
+
+    const resolutionSelect = affectsTableRows[0].find('td:nth-of-type(6) span');
+    expect(resolutionSelect.text()).toBe('');
```
MrMarble, line 253: Could you add another expect at the beginning for the resolution? How do we know that it wasn't always `''`?

C-Valen: Makes sense, added.

src/components/FlawAffects/__tests__/FlawAffects.spec.ts
```diff
@@ -233,6 +233,30 @@ describe('flawAffects', () => {
...
+    const affectednessSelect = affectsTableEditingRows[0].find('td:nth-of-type(5) select');
+    await affectednessSelect.setValue('NOTAFFECTED');
+    const selectedAffectedness = affectednessSelect.find('option[selected]');
+    expect(selectedAffectedness.text()).toBe('NOTAFFECTED');
+
+    const affectRowCommitBtn = affectsTableEditingRows[0].find('td:last-of-type button:first-of-type');
+    expect(affectRowCommitBtn.exists()).toBe(true);
+    await affectRowCommitBtn.trigger('click');
+
+    resolutionSelect = affectsTableRows[0].find('td:nth-of-type(6) span');
```
MrMarble, line 256: I think there's no need to reassign the element again, as this point to the DOM element that should be updated

C-Valen: I think it is necessary, even if the DOM is updated the variable declared is a constant.
Removing last assignment makes the expect after it to fail.

src/components/FlawAffects/__tests__/FlawAffects.spec.ts
```diff
@@ -233,6 +233,30 @@ describe('flawAffects', () => {
     expect(resolutionOptions.includes('DEFER')).toBe(false);
   });
 
+  osimFullFlawTest('Changing affect to not affected automatically set\'s empty resolution', async () => {
```
superbuggy, line 236:
```suggestion
  osimFullFlawTest('Changing affect to not affected automatically sets empty resolution', async () => {
```

src/components/FlawAffects/FlawAffectsTableRow.vue
```diff
@@ -188,13 +188,13 @@ function affectednessChange(event: Event, affect: ZodAffectType) {
         v-model="affect.affectedness"
         class="form-select"
         @keydown="handleKeystroke($event, affect)"
+        @change="affectednessChange($event, affect)"
```
superbuggy, line 191: Think this is probably fine, but wanted to mention using `watch` is another option in case props/model would change for some reason.

superbuggy (changes requested):
Requesting one language change

src/components/FlawAffects/__tests__/FlawAffects.spec.ts
MrMarble: The variable being a constant only means it cannot be reassigned, but if the value is a reference it can update on its own, but you're right, the VueWrapper does not update on its own, you have to query it manually https://github.com/vuejs/test-utils/issues/1106

MrMarble (approved):

MrMarble (approved):

superbuggy (approved):

PR #532, opened 2025-01-27
✨ Make per page number an input field

C-Valen:
# OSIDB-3507 Make per page number an input field

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Converts the number of items per page indicator in the affects and trackers table to an input field so the users can set a custom for the number of rows displayed in each table.

## Changes:

- Converts items per page indicator to an input field
- Makes default per page range limits wider (1 - 100)

## Demo:

https://github.com/user-attachments/assets/bbeb27fc-69d3-460d-9b39-0768ffaf281e

## Considerations:

- Users are still able to use the increase/decrease buttons to change that value

Closes OSIDB-3507


src/components/FlawTrackers/FlawTrackers.vue
```diff
@@ -358,7 +365,10 @@ const tableTrackers = computed(() => {
       .per-page-btn {
         display: flex;
         padding-block: 0;
-        pointer-events: none;
+
+        input {
+          width: 3rem;
```
superbuggy, line 370: could use `ch` also

superbuggy (approved):

MrMarble (approved):

PR #559, opened 2025-03-25
Feature/osidb 2363 cvss4 support

superbuggy:
# [OSIDB-ID] [Title]

## Checklist:

- [ ] Commits consolidated
- [ ] Changelog updated
- [ ] Jira ticket updated

## Summary:

[Replace with a brief summary of the changes introduced by this PR.]

## Changes:

## Considerations:


C-Valen (changes requested):
I have been reviewing the code here and testing the UI on UAT But I believe there are some discrepancies between both. Probably this PR is not up to date with UAT?

Some considerations:
- In some cases the synchronization of RH vectors does not work
- If we are keeping the overlayed display design I think we should make the NVD Vector field (and probably all the collapsible div if extended) visible.
A possible solution can be to display the calculator over the right side of the flaw form.
- Affect's calculator looks a bit buggy now
- RH CVSSv4 Is displayed in the collapsible of Non-RH CVSS
- The CVSSv4 by default has a vector when creating a new flaw (instead of being empty)
image

superbuggy:
> I have been reviewing the code here and testing the UI on UAT But I believe there are some discrepancies between both. Probably this PR is not up to date with UAT?

I can merge in the code for OSIDB-4197, that may make this easier to review after all

src/components/CvssCalculator/Cvss3Calculator/Cvss3Calculator.vue
```diff
@@ -9,43 +9,44 @@ import {
...
 
 function updateFactors(newCvssVector: null | string | undefined) {
-  if (cvssVector.value !== newCvssVector) {
-    cvssVector.value = newCvssVector;
-  }
   cvssFactors.value = getFactors(newCvssVector ?? '');
+  if (props.cvssVector !== newCvssVector) {
+    emit('update:cvssVector', newCvssVector ?? null);
+  }
 }
```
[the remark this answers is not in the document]
superbuggy: This is sort of a quirk, though not ideal, of the new logic. The factors are mostly used in the CVSS3 calculator composable and not in the giant singleton CVSS4 class, so instead of just having a watcher handle the cvssFactor changes, I am using a specific function to update the vector here since that's what works for the CVSS4 class.

src/components/CvssCalculator/CvssCalculator.vue
```diff
@@ -65,7 +66,8 @@ function handlePaste(e: ClipboardEvent) {
   }
 
   updateFactors(formatFactors(cvssFactors.value));
-  cvssScore.value = calculateScore(cvssFactors.value);
+  updateScore(calculateScore(cvssFactors.value) ?? 0);
+  // cvssScore.value = calculateScore(cvssFactors.value);
```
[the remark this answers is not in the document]
superbuggy: 🔥

superbuggy:
> Some considerations:
> 
> * In some cases the synchronization of RH vectors does not work

Curious to know what you noticed with this

> * If we are keeping the overlayed display design I think we should make the NVD Vector field (and probably all the collapsible div if extended) visible.
>   A possible solution can be to display the calculator over the right side of the flaw form.
> * Affect's calculator looks a bit buggy now

Maybe we can chat about these during a code review and coordinate the UI changes.

> * RH CVSSv4 Is displayed in the collapsible of Non-RH CVSS
> * The CVSSv4 by default has a vector when creating a new flaw (instead of being empty)

These should be fixed now, great catches--thank you.

MrMarble (changes requested):
I have not finished reading everything, but I have already some comments that needs changes

src/components/CvssCalculator/Cvss3Calculator/Cvss3Calculator.vue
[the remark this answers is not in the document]
superbuggy: The CVSS4 factors dont update right away, thanks for demoing @C-Valen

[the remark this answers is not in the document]
superbuggy: I believe this is resolved

src/components/CvssCalculator/Cvss4Calculator/Cvss4Calculator.vue
```diff
@@ -0,0 +1,156 @@
...
+  emit('update:cvssScore', newCvss4Score ?? null);
+}
+
+function updateCvss4Vector(newCvss4Vector: null | string | undefined) {
+  emit('update:cvssVector', newCvss4Vector ?? null);
+}
+
+// TODO: Move these into composable?
+watch(() => props.cvss4Score, updateCvss4Score);
+watch(() => props.cvss4Vector, updateCvss4Vector);
```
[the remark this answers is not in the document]
superbuggy: Oh this is a leftover task, this is cleaned up now.

src/components/CvssCalculator/CvssCalculator.vue
```diff
@@ -1,40 +1,40 @@
...
-const cvssVector = defineModel<null | string | undefined>('cvssVector');
-const cvssScore = defineModel<null | number | undefined>('cvssScore');
-
-const error = computed(() => validateCvssVector(cvssVector.value));
-const cvssFactors = ref<Record<string, string>>({});
+import Cvss3Calculator from '@/components/CvssCalculator/Cvss3Calculator/Cvss3Calculator.vue';
+import Cvss4Calculator from '@/components/CvssCalculator/Cvss4Calculator/Cvss4Calculator.vue';
+
+import { useCvssScores, validateCvssVector } from '@/composables/useCvssScores';
+import { getFactors, calculateScore, formatFactors } from '@/composables/useCvss3Calculator';
```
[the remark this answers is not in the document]
superbuggy: Agreed, this is confusing

src/components/CvssCalculator/CvssCalculator.vue
```diff
@@ -1,40 +1,40 @@
...
+  cvss4Selections,
+  cvss4Vector,
+  cvssFactors,
+  cvssScore,
+  cvssVector,
+  cvssVersion,
+  parseVectorV4String,
+  setMetric,
+  shouldSyncCvssFactors,
+  updateCvss3Factors,
```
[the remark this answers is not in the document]
superbuggy: It is scattered, it's a consequence of the pattern some of the older components having update logic that probably belonged in the CVSS3 composable, I can probably pay off this tech debt without much trouble.

superbuggy:
> Second review form my side is in progress but wanted to advance this issue.
> 
> CVSS4 vector is initialized with some default factors. If you open a flaw that doesn't have CVSS4 it will show a vector filled in.
> 
> image
> 
> Edit: I think this could be the source of other bugs I am noticing in CVSSv4 (e.g. Erase button doesn't clear vector, but restores default one, version synchronization not working both sides, ...)

Ah okay, I am noticing the sync breaks from V4 to V3 when you choose a value already present in the default CVSS4 vector. I  thought of a fix for that. Unfortunately that the default CVSS4 vector is inevitable in the hard-forked CVSS4 code. The fix will have to go on top of that code to avoid modifying it.

src/components/CvssCalculator/CvssCalculator.vue
```diff
@@ -1,40 +1,40 @@
...
+import RedHatIconSvg from '@/assets/Logo-Red_Hat-Hat_icon-Standard-RGB.svg';
+import { CvssVersions, CvssVersionDisplayMap } from '@/constants';
+
+const {
+  cvss3Factors,
+  cvss4Score,
+  cvss4Selections,
+  cvss4Vector,
+  cvssScore,
+  cvssVector,
```
C-Valen, line 20: Are these two referring to v3?
If so I suggest adding the version number to the name, cause at some point they wont be the default RH CVSS version.

src/components/CvssCalculator/CvssCalculator.vue
```diff
@@ -88,24 +95,55 @@ function highlightFactorValue(factor: null | string) {
...
+          </span>
         </span>
         <div class="input-wrapper col">
           <CvssVectorInput
             ref="cvssVectorInput"
-            :cvssVector="cvssVector"
-            :cvssScore="cvssScore"
-            :cvssFactors="cvssFactors"
+            :cvss4Vector="cvss4Vector"
+            :cvss3Factors="cvss3Factors"
```
C-Valen, line 136: Personal taste suggestion (also applies for other cases)
```suggestion
            :cvss4Vector
            :cvss3Factors
```

src/components/CvssCalculator/CvssCalculatorOverlayed.vue
```diff
@@ -2,42 +2,42 @@
...
+const props = defineProps<{
+  affect: ZodAffectType;
 }>();
 
-const error = computed(() => validateCvssVector(cvssVector.value));
-const cvssFactors = ref<Record<string, string>>({});
+const { cvss3Factors, cvssScore, cvssVector, cvssVersion, updateScore, updateVector } = useCvssScores(props.affect);
+
+const error = computed(() => validateCvssVector(cvssVector.value, cvssVersion.value) ?? null);
+// const cvss3Factors = ref<Record<string, string>>({});
```
C-Valen, line 24: still need it?

C-Valen (changes requested):
Possible bugs - Tested on `CVE-2001-0414 UAT`
- Default CVSSv4 has non empty vector
- Changing a new affect CVSS changes flaw's CVSS
- Erase button doesn't work properly on v4
- Cannot deselect a selected factor by clicking again on it on v4

++ On the new v4 design, nice work there.
I'd reduce height a bit more removing some vertical space.

Will continue looking at the code tomorrow.

src/components/CvssCalculator/CvssCalculator.vue
superbuggy: they refer to the currently selected cvss

superbuggy:
> Ah okay, I am noticing the sync breaks from V4 to V3 when you choose a value already present in the default CVSS4 vector. I thought of a fix for that. Unfortunately that the default CVSS4 vector is inevitable in the hard-forked CVSS4 code. The fix will have to go on top of that code to avoid modifying it.

There was not a way to fix this without modifying the underlying class. Making this modification without breaking anything took and tedious careful experimentation because the calculation logic 1) *relies on hardcoded default values* and  2) *does not expect there to be unfilled values*.

superbuggy:
> Possible bugs - Tested on `CVE-2001-0414 UAT`
> 
> * Default CVSSv4 has non empty vector
> * Changing a new affect CVSS changes flaw's CVSS
> * Erase button doesn't work properly on v4
> * Cannot deselect a selected factor by clicking again on it on v4

These missing features and bugs have been addressed

C-Valen (commented):
Previous comment points are addressed, good work.

I noticed the erase button always removes both version vectors even if the synchronization is disabled. Would it make sense to make it independent in such case?

Also the scores are not reset when erasing multiple versions.

https://github.com/user-attachments/assets/44b77d0c-7619-4fe8-9456-f5c1f60b0903

src/components/ChangeLog/ChangeLog.vue
```diff
@@ -13,7 +13,9 @@ const { closeModal, isModalOpen, openModal } = useModal();
 const html = ref({ header: '', body: '' });
 
 function fetchChangeLog() {
-  return fetch('/CHANGELOG.md')
+  const scriptEl = document.head.querySelector('script[src]') as HTMLScriptElement;
+  const buildString = scriptEl?.src?.match(/\/index-(\w+)\.js/)?.[1];
+  return fetch(`/CHANGELOG.md?build=${buildString}`)
```
MrMarble, line 18: Oh Nice! 💯 
I thought about using Vite build process to add the hash, but this works

src/composables/useCvss4Calculator.ts
```diff
@@ -0,0 +1,90 @@
...
+  const cvss4Selections = ref(
+    deepMap(value => Array.isArray(value) ? null : value, METRICS),
+  );
+  const cvss4Vector = computed(() => {
+    const vector = Object.values(cvss4Selections.value).flatMap((group: any) =>
+      Object.entries(group)
+        .filter(([, value]) => value !== null)
+        .map(([metric, value]) => `${metric}:${value}`),
+    ).join('/');
+    return `CVSS:4.0/${vector}`;
```
MrMarble, line 41: This is setting the string `CVSS:4.0/` always, even if the vector is empty, this seems to introduce a werid behavior in the component, as jumping from an empty CvssV3 to a V4, now suddenly the vector is invalid, and since is the default behavior, there's no way of clearing the input, thankfully the V3 can be set to 0, if not the user would be stuck with an invalid flaw until a random vector is typed or the tab reloaded.

Also for some reason the V4 calculator sets an score of 1 when the input is empty




https://github.com/user-attachments/assets/6df93d6a-8b76-469f-b880-cc8ca5cd5ef4

src/components/CvssCalculator/CvssVectorInput.vue
```diff
@@ -62,11 +62,11 @@ const getFactorColor = (weight: number, isHovered: boolean = false) => {
       class="osim-cvss-score"
       :style="isFocused ? {color: 'white'} : {color: 'black'}"
     >
-      {{ cvssScore != null ? `${cvssScore} ` : '' }}
+      {{ cvssScore !== null ? `${cvssScore} ` : '' }}
```
MrMarble, line 65: I wonder if this strict null check could be problematic...
I tried to follow the usage and seems to be handled, the only non null expression (I found) is this one:
https://github.com/RedHatProductSecurity/osim/blob/b24b26dd6b943ae346381fada87d6b9b3f2feff8/src/composables/useCvssScores.ts#L152
In the case the `rhCvss` is not a ref ordoes not have `score`, is going to return `undefined`, hopefully this should not happen, as the shape ofthe `rhCvss` is hardcoded, but we should keep it in mind.

We could use the nullish coalescing there as well
```js
const cvssScore = computed(() => rhCvss.value?.score ?? null); 
```

MrMarble (approved):

src/utils/__tests__/cvss40.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+        : { vector: vectorScorePair[0], score: Number.parseFloat(vectorScorePair[1]) };
+    }).filter(Boolean);
+    data[file.name] = scoredVectors;
+    return data;
+  }, {} as Record<string, any>);
+
+  Object.entries(testData).forEach(
+    ([fileName, vectorScores]: [string, any]) => {
+      it(`should calculate scores in ${fileName} correctly`, () => {
+        vectorScores.slice(0, 1000).forEach(({ score, vector }: { score: number; vector: string }) => {
```
MrMarble, line 277: Why do we have 500.000 vectors if we only test the first 1000 on each file?

PR #560, opened 2025-03-25
🐛 Move privacyNotice cookie to settings store

C-Valen:
# OSIDB-4129 Move privacyNotice cookie to settings store

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Refactors the use of localsotrage for the privacyNotice cookie to use the setting store of the user to fix issues with cache variables (API keys values getting overrived)

## Changes:

- Move privacyNotice cookie into settingsStore
- Update tests

## Considerations:

- Couldn't replicate the issue so will need to get feedback from using on UAT

Closes OSIDB-4129

### DIFF
--- a/src/stores/SettingsStore.ts
+++ b/src/stores/SettingsStore.ts
@@ -1,8 +1,8 @@
-import { ref, watch } from 'vue';
-
 import { defineStore } from 'pinia';
 import { z } from 'zod';
-import { useStorage } from '@vueuse/core';
+import { useLocalStorage } from '@vueuse/core';
+
+import { useToastStore } from './ToastStore';
 
 export const SettingsSchema = z.object({
   // bugzillaApiKey: z.string().length(
@@ -14,6 +14,7 @@ export const SettingsSchema = z.object({
   affectsPerPage: z.number(),
   trackersPerPage: z.number(),
   isHidingLabels: z.boolean().optional().default(false),
+  privacyNoticeShown: z.boolean().default(false),
 });
 
 export type SettingsType = z.infer<typeof SettingsSchema>;
@@ -25,12 +26,13 @@ const defaultValues: SettingsType = {
   affectsPerPage: 10,
   trackersPerPage: 10,
   isHidingLabels: false,
+  privacyNoticeShown: false,
 };
-const osimSettings = useStorage('OSIM::USER-SETTINGS', structuredClone(defaultValues));
 
 export const useSettingsStore = defineStore('SettingsStore', () => {
-  const settings = ref<SettingsType>(osimSettings.value);
-  // const settings = useSessionStorage(_settingsStoreKey, {} as SettingsType);
+  const { addToast } = useToastStore();
+
+  const settings = useLocalStorage('OSIM::USER-SETTINGS', structuredClone(defaultValues));
 
   const validatedSettings = SettingsSchema.safeParse(settings.value);
   if (validatedSettings.success) {
@@ -41,12 +43,19 @@ export const useSettingsStore = defineStore('SettingsStore', () => {
     settings.value = structuredClone(defaultValues);
   }
 
-  watch(settings, () => {
-    osimSettings.value = settings.value;
-  });
+  if (!settings.value.privacyNoticeShown) {
+    addToast({
+      title: 'Privacy Notice',
+      body: 'OSIM transmits input information externally to Bugzilla for the purpose of retrieving bug data.' +
+      ' In some cases that information may be publicly visible.',
+    });
+    settings.value.privacyNoticeShown = true;
+  }
 
-  function save(newSettings: SettingsType) {
-    settings.value = { ...newSettings };
+  // This is to remove previous privacy notice shown flag
+  // Can be removed in future versions, when we are sure that all users have updated to the latest version
+  if (localStorage.getItem('privacyNoticeShown')) {
+    localStorage.removeItem('privacyNoticeShown');
   }
 
   function $reset() {
@@ -55,7 +64,6 @@ export const useSettingsStore = defineStore('SettingsStore', () => {
 
   return {
     $reset,
-    save,
     settings,
   };
 });
--- a/src/stores/__tests__/ToastStore.spec.ts
+++ b/src/stores/__tests__/ToastStore.spec.ts
@@ -2,22 +2,15 @@ import { createPinia, setActivePinia } from 'pinia';
 
 import { useToastStore, type ToastNew } from '@/stores/ToastStore';
 
-import { useSettingsStore } from '../SettingsStore';
-
 describe('toastStore', () => {
   let toastStore: ReturnType<typeof useToastStore>;
-  let settingsStore: ReturnType<typeof useSettingsStore>;
 
   beforeEach(() => {
     setActivePinia(createPinia());
-    settingsStore = useSettingsStore();
     toastStore = useToastStore();
   });
 
   it('addToast', () => {
-    settingsStore.settings.privacyNoticeShown = true;
-    settingsStore.settings.showNotifications = true;
-
     const toast: ToastNew = { body: 'Test Toast' };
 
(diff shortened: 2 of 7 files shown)


src/components/Navbar/Navbar.vue
```diff
@@ -61,6 +61,17 @@ const usernameDisplay = computed(() => {
...
+
+onMounted(() => {
+  if (!settings.privacyNoticeShown) {
+    addToast({
+      title: 'Privacy Notice',
+      body: 'OSIM transmits input information externally to Bugzilla for the purpose of retrieving bug data.' +
+      ' In some cases that information may be publicly visible.',
+    });
+    settings.privacyNoticeShown = true;
+  }
```
superbuggy, line 73: Maybe move this to `SettingsStore.ts`

src/stores/SettingsStore.ts
```diff
@@ -25,6 +26,7 @@ const defaultValues: SettingsType = {
   affectsPerPage: 10,
   trackersPerPage: 10,
   isHidingLabels: false,
+  privacyNoticeShown: false,
 };
 const osimSettings = useStorage('OSIM::USER-SETTINGS', structuredClone(defaultValues));
```
superbuggy, line 32: Potentially we should try to remove the extra settings ref (replace `osimSettings` with just `settings` outside the store) and just modify that directly

superbuggy: that way we could remove the unused save function and the watcher along with the settings ref

src/stores/__tests__/ToastStore.spec.ts
```diff
@@ -15,13 +15,14 @@ describe('toastStore', () => {
   });
 
   it('addToast', () => {
```
superbuggy, line 17:
```suggestion
  it('shows privacy notice toast', () => {
```

src/stores/__tests__/ToastStore.spec.ts
```diff
@@ -15,13 +15,14 @@ describe('toastStore', () => {
...
+    settingsStore.settings.privacyNoticeShown = true;
```
MrMarble, line 18:
```suggestion
```
This change is not really needed, the `addToast` call in https://github.com/RedHatProductSecurity/osim/blob/a106a274b0932e83b47f0b588b29b7772bc15a7b/src/stores/SettingsStore.ts#L46-L47 only happens on store initialization, and since it is initialized on the `beforeEach` is doing nothing.

> Actually since the store is persisted in localStorage, is affecting the next test but not itself, if you change the value to `false` the `reset` test will fail.

MrMarble: On second thought, the ToastStore does need the settings store at all, we are **actually adding a side effect on the tests by using it**, we should remove it entirely to keep the tests as a single unit

```diff
 import { useToastStore, type ToastNew } from '@/stores/ToastStore';
 
-import { useSettingsStore } from '../SettingsStore';

 describe('toastStore', () => {
   let toastStore: ReturnType<typeof useToastStore>;
-  let settingsStore: ReturnType<typeof useSettingsStore>;
 
   beforeEach(() => {
     setActivePinia(createPinia());
-    settingsStore = useSettingsStore();
     toastStore = useToastStore();
   });
 
   it('addToast', () => {
-    settingsStore.settings.privacyNoticeShown = true;
-    settingsStore.settings.showNotifications = true;
-
     const toast: ToastNew = { body: 'Test Toast' };
 
     toastStore.addToast(toast);
 
-    expect(toastStore.toasts).toHaveLength(2);
+    expect(toastStore.toasts).toHaveLength(1);
     expect(toastStore.toasts[0].isFresh).toBe(true);
   });
```

src/stores/SettingsStore.ts
```diff
@@ -25,12 +26,13 @@ const defaultValues: SettingsType = {
...
+  privacyNoticeShown: false,
 };
-const osimSettings = useStorage('OSIM::USER-SETTINGS', structuredClone(defaultValues));
 
 export const useSettingsStore = defineStore('SettingsStore', () => {
-  const settings = ref<SettingsType>(osimSettings.value);
-  // const settings = useSessionStorage(_settingsStoreKey, {} as SettingsType);
+  const { addToast } = useToastStore();
+
+  const settings = useLocalStorage('OSIM::USER-SETTINGS', structuredClone(defaultValues));
```
MrMarble, line 35: >Should we use a pinia plugin for persisting data? 
We are handling this differently on each store, maybe is worth centralizing the logic to one place.
We could write our own plugin or use the community one https://github.com/prazdevs/pinia-plugin-persistedstate

src/stores/__tests__/ToastStore.spec.ts
MrMarble: please read my comments on https://github.com/RedHatProductSecurity/osim/pull/560#discussion_r2042212502

MrMarble:
Reading the user statements on the problem I don't see how this fixes the problem, maybe I'm wrong, here's how I understand it:

When the user has multiple tabs open, updating the token in one of the tabs does not propagate the change to the others, which still uses the old token, any change to the settings will override the stored token before it gets a change to get loaded again in memory.

Currently we don't support multiple tabs, but since the users keep opening multiple tabs, maybe we should fix this?

src/stores/SettingsStore.ts
```diff
@@ -41,12 +43,19 @@ export const useSettingsStore = defineStore('SettingsStore', () => {
     settings.value = structuredClone(defaultValues);
   }
 
-  watch(settings, () => {
-    osimSettings.value = settings.value;
-  });
+  if (!settings.value.privacyNoticeShown) {
+    addToast({
```
MrMarble, line 47: I'll add a new test in the `SettingsStore.spec.ts` file for this:

```typescript
vi.mock('@/stores/ToastStore', () => {
  const addToast = vi.fn();
  return ({
    useToastStore: vi.fn(() => ({ addToast })),
  });
});

// ...

  it('shows privacyNotification on initialization', async () => {
    const { addToast } = vi.mocked(useToastStore());

    expect(settingsStore.settings.privacyNoticeShown).toEqual(true);
    expect(addToast).toHaveBeenCalledOnce();
    expect(addToast).toHaveBeenCalledWith(expect.objectContaining({ title: 'Privacy Notice' }));
  });

  it('only shows privacyNotification once', async () => {
    const { addToast } = vi.mocked(useToastStore());

    settingsStore.settings.privacyNoticeShown = false;
    await flushPromises();
    // reset pinia instance so we read from the storage instead of the memory
    settingsStore = useSettingsStore(createPinia());

    expect(addToast).toHaveBeenCalledOnce();
  });
```
> ToastStore imports need to be updated for mock to work

src/stores/SettingsStore.ts
superbuggy: we can leave it in the store since that's basically global anyway

src/stores/__tests__/ToastStore.spec.ts
C-Valen: As discussed, agree to have the PrivacyNotice test under `SettingsStore` suite

src/stores/SettingsStore.ts
C-Valen: I was having issues dealing with these tests. By some reason the `addToast` mock method is being called but the test doesn't interpret it as such.

To don't delay more this fix release I will add a integration test that covers the same case of showing the privacy notice on user's first login.

C-Valen: [Privacy Notice Integration Test](https://github.com/RedHatProductSecurity/osim-ui-tests/pull/4/commits/649abfe9951aa69c52a090534bd3d646a33454db)

MrMarble (dismissed):

MrMarble (approved):

superbuggy (approved):

PR #570, opened 2025-05-16
✨ Comments section display modes

C-Valen:
# OSIDB-3467 Comments section display modes

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Adding a new display mode to show all comment types under the same view. We are keeping previous display mode and allow users to choose their preference.

## Changes:

- Refactor `FlawComments` into smaller components for simplicity
- Relocate some comments constants/types to be globally available
- Add new single view for all comment types
- Support toggling between comments view modes
- Order comments by date descendant
- Use local storage for user display mode selection
- Make comments toolbar sticky
- Update tests

## Demo
- Working on it, (having problems with Jira stage account)

## Considerations:

- Comment toolbar sticky is mostly intended to allow users to use the view mode switch at any scroll point of the section 
- NTH - Custom comments order

Closes OSIDB-3467

### DIFF
--- a/src/components/FlawComments/CommentsToolbar.vue
+++ b/src/components/FlawComments/CommentsToolbar.vue
@@ -78,7 +78,7 @@ const { settings } = storeToRefs(useSettingsStore());
     <div class="form-check form-switch ms-auto" style="font-size: 18px">
       <input
         id="singleViewSwitch"
-        v-model="settings['singleCommentsView']"
+        v-model="settings['unifiedCommentsView']"
         class="form-check-input"
         type="checkbox"
       >
--- a/src/components/FlawComments/CommentsToolbar.vue
+++ b/src/components/FlawComments/CommentsToolbar.vue
@@ -1,5 +1,8 @@
 <script setup lang="ts">
+import { storeToRefs } from 'pinia';
+
 import { jiraTaskUrl } from '@/services/JiraService';
+import { useSettingsStore } from '@/stores/SettingsStore';
 import { CommentType } from '@/types';
 
 defineProps<{
@@ -12,13 +15,13 @@ defineProps<{
   showBugzillaLink: boolean;
   showJiraLink: boolean;
   taskKey: null | string | undefined;
-  unifiedView: boolean;
 }>();
 
 const emit = defineEmits<{
   startNewComment: [type: CommentType];
-  toggleView: [];
 }>();
+
+const { settings } = storeToRefs(useSettingsStore());
 </script>
 
 <template>
@@ -74,11 +77,9 @@ const emit = defineEmits<{
     </div>
     <div class="form-check form-switch ms-auto" style="font-size: 18px">
       <input
-        id="singleViewSwitch"
+        v-model="settings['singleCommentsView']"
         class="form-check-input"
         type="checkbox"
-        :value="unifiedView"
-        @click="emit('toggleView')"
       >
       <label class="form-check-label" for="singleViewSwitch" style="user-select: none;">Unified view</label>
     </div>
--- a/src/components/FlawComments/CommentsToolbar.vue
+++ b/src/components/FlawComments/CommentsToolbar.vue
@@ -1,0 +1,86 @@
+<script setup lang="ts">
+import { jiraTaskUrl } from '@/services/JiraService';
+import { CommentType } from '@/types';
+
+defineProps<{
+  bugzillaLink: string;
+  internalCommentsAvailable: boolean;
+  isAddingNewComment: boolean;
+  isSaving: boolean;
+  newCommentAllowed: boolean;
+  newCommentType: CommentType;
+  showBugzillaLink: boolean;
+  showJiraLink: boolean;
+  taskKey: null | string | undefined;
+  unifiedView: boolean;
+}>();
+
+const emit = defineEmits<{
+  startNewComment: [type: CommentType];
+  toggleView: [];
+}>();
+</script>
+
+<template>
+  <div class="d-flex mb-4 sticky-top p-3 ps-0 border-bottom align-items-center" style="background-color: white;">
+    <div class="d-flex gap-2">
+      <button
+        type="button"
+        class="btn btn-secondary"
+        :disabled="!newCommentAllowed || isSaving || (isAddingNewComment && newCommentType === CommentType.Public)"
+        @click="emit('startNewComment', CommentType.Public)"
+      >
+        Add Public Comment
+      </button>
+      <button
+        type="button"
+        class="btn btn-secondary"
+        :disabled="!newCommentAllowed || isSaving || (isAddingNewComment && newCommentType === CommentType.Private)"
+        @click="emit('startNewComment', CommentType.Private)"
+      >
+        Add Private Comment
+      </button>
+      <button
+        type="button"
+        class="btn btn-secondary"
+        :disabled="
+          !internalCommentsAvailable
+            || !newCommentAllowed
+            || isSaving
+            || (isAddingNewComment && newCommentType === CommentType.Internal)
+        "
+        @click="emit('startNewComment', CommentType.Internal)"
+      >
+        Add Internal Comment
+      </button>
+      <a
+        v-if="showBugzillaLink"
+        :href="bugzillaLink"
+        target="_blank"
+        class="btn tab-btn border border-secondary"
+        :disabled="isSaving"
+      >
+        View in Bugzilla
+      </a>
+      <a
+        v-if="showJiraLink"
+        :href="taskKey ? jiraTaskUrl(taskKey) : '#'"
+        target="_blank"
+        class="btn tab-btn border border-secondary"
+        :disabled="isSaving"
+      >
+        View in Jira
+      </a>
+    </div>
+    <div class="form-check form-switch ms-auto" style="font-size: 18px">
+      <input
+        id="singleViewSwitch"
+        class="form-check-input"
+        type="checkbox"
+        :value="unifiedView"
+        @click="emit('toggleView')"
+      >
+      <label class="form-check-label" for="singleViewSwitch" style="user-select: none;">Unified view</label>
+    </div>
+  </div>
+</template>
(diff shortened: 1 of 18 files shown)


src/components/__tests__/FlawComments.spec.ts
```diff
@@ -14,11 +16,14 @@ vi.mock('@/services/JiraService', () => ({
 }));
 
 describe('flawComments', () => {
+  let settingsStore: ReturnType<typeof useSettingsStore>;
```
MrMarble, line 19: Shouldn't this warning be an error?

>'settingsStore' is assigned a value but never used. Allowed unused vars must match /^_/u

src/composables/useFlawCommentsModel.ts
```diff
@@ -3,7 +3,8 @@ import { computed, ref, type Ref } from 'vue';
 import { getJiraComments, postJiraComment } from '@/services/JiraService';
 import type { ZodFlawCommentType, ZodFlawType } from '@/types/zodFlaw';
 import { postFlawComment } from '@/services/FlawService';
-import { CommentType, SYSTEM_EMAIL } from '@/constants';
+import { SYSTEM_EMAIL } from '@/constants';
+import type { CommentType } from '@/constants';
```
MrMarble, line 7:
```suggestion
import { SYSTEM_EMAIL, type CommentType } from '@/constants';

```

src/constants/index.ts
```diff
@@ -42,3 +35,17 @@ export const allowedSources = [
...
+  Private,
+  Internal,
+  System,
+}
+
+export const commentTooltips: Record<CommentType, string> = {
+  [CommentType.Public]: 'Bugzilla Public - This comments are visible to everyone.',
+  [CommentType.Private]: 'Bugzilla Private - This comments are visible to Red Hat associates.',
+  [CommentType.Internal]: 'Jira Internal - This comments are visible to team members with required permissions.',
+  [CommentType.System]: 'Bugzilla System - This are auto-generated private comments.',
```
MrMarble, line 50:
```suggestion
export const commentTooltips: Record<CommentType, string> = {
  [CommentType.Public]: 'Bugzilla Public - These comments are visible to everyone.',
  [CommentType.Private]: 'Bugzilla Private - These comments are visible to Red Hat associates.',
  [CommentType.Internal]: 'Jira Internal - These comments are visible to team members with required permissions.',
  [CommentType.System]: 'Bugzilla System - These are auto-generated private comments.',
```

src/components/FlawComments/FlawComments.vue
```diff
@@ -77,181 +79,91 @@ onMounted(async () => {
...
+  if (props.publicComments.includes(comment)) {
+    return CommentType.Public;
+  } else if (props.privateComments.includes(comment)) {
+    return CommentType.Private;
+  } else if (props.internalComments.includes(comment)) {
+    return CommentType.Internal;
+  } else {
+    return CommentType.System;
+  }
 }
```
MrMarble, line 107: I think this could be moved to the `useFlawComments` composable

src/components/FlawComments/FlawComments.vue
```diff
@@ -77,181 +79,91 @@ onMounted(async () => {
...
 
-function parseJiraTags(text: string) {
-  const jiraUserLink = `${osimRuntime.value.backends.jiraDisplay}/ViewProfile.jspa?name=`;
-  const jiraTagRegex = /\[~([^[\]]+)\]/g;
-
-  return text
-    .replace(jiraTagRegex, (match, p1) => `<a target="_blank" href="${jiraUserLink}${p1}">${p1}</a>`);
+function getCommentListWithTypes(list: ZodFlawCommentType[]) {
+  return list.map(c => ({ ...c, type: getCommentType(c) }));
```
MrMarble, line 110: I think I would prefer to add the `type` directly in Zod and change the filter functions used in `useFlawComments` to refine function in Zod, we have all the data we need there.

src/components/FlawComments/FlawComments.vue
```diff
@@ -46,28 +43,33 @@ const commentLabels = computed(() => {
...
+      return orderCommentsByDate(props.privateComments);
     case CommentType.Internal:
-      return props.internalComments;
+      return orderCommentsByDate(props.internalComments);
     case CommentType.System:
-      return props.systemComments;
+      return orderCommentsByDate(props.systemComments);
     default:
       return [];
   }
```
MrMarble, line 75: If we sort the comments directly in `useFlawComments` we could have a map of `CommentType` to comments array.
```typescript
const commentsByType = {
  [CommnetType.Pulblic]: publicComments,
  [CommnetType.Private]: privateComments,
  [CommnetType.Internal]: internalComments,
  [CommnetType.System]: systemComments,
}
```

using the comments then is just:
```vue
<CommentList :commentList="commentsByType[selectedTab]" /> 
```

This way we can remove the computed property, I don't think there's any performance difference (there's some, as is one computed ref less, and direct access to an in memory dictionary), and this `commentsByType` map could be part of the `useFlawComments` composable as well

src/composables/useFlawCommentsModel.ts
MrMarble, line 1: What do you think about taking the change to remove some of the clutter from `useFlawModel`?
I think we could remove the call to `useFlawCommentsModel`, directly calling it from the `FlawComments` component, we will need to pass the `isSaving` and `afterSaveSuccess` callback for now until we extract them out, but will help with the future cleaning.
There's some logic in the `FlawComments` component that I think it makes sense to be inside the composable, but since we are using the composable through the useFlawModel, using props, is a bit cumbersome

src/components/FlawComments/CommentsToolbar.vue
```diff
@@ -0,0 +1,88 @@
...
+      </a>
+    </div>
+    <div class="form-check form-switch ms-auto" style="font-size: 18px">
+      <input
+        id="singleViewSwitch"
+        v-model="settings['singleCommentsView']"
+        class="form-check-input"
+        type="checkbox"
+      >
+      <label class="form-check-label" for="singleViewSwitch" style="user-select: none;">Single view</label>
```
MrMarble, line 85: Just a nitpick, I think `unified` is a better word for this
```suggestion
      <label class="form-check-label" for="singleViewSwitch" style="user-select: none;">Unified view</label>
```

> Google AI output: A unified view refers to presenting data from multiple sources as a single, coherent whole.

MrMarble (commented):
Some minor comments, overall looks good!

src/composables/useFlawCommentsModel.ts
C-Valen: I'd prefer leaving this for now, this PR has already changes with quite impact to adopt a new refactor in the  high level structure of the composables.

src/components/FlawComments/FlawComments.vue
```diff
@@ -46,212 +40,90 @@ const commentLabels = computed(() => {
...
-  }
-};
+const newCommentAllowed = computed(() =>
+  (
+    (
+      (
+        selectedTab.value !== CommentType.System
+        && (props.internalCommentsAvailable || selectedTab.value !== CommentType.Internal)
+      )
+      || (settings.value['singleCommentsView']))
```
MrMarble, line 78:
```suggestion
      || (settings.value['unifiedCommentsView']))
```

src/components/FlawComments/FlawComments.vue
```diff
@@ -46,212 +40,90 @@ const commentLabels = computed(() => {
...
+      :taskKey
+      :isAddingNewComment
+      :isSaving
+      :newCommentType
+      :internalCommentsAvailable
+      @saveComment="(newComment) => handleCommentSave(newComment)"
+      @cancelComment="() => isAddingNewComment = false"
+    />
     <Tabs
+      v-if="!settings['singleCommentsView']"
```
MrMarble, line 121:
```suggestion
      v-if="!settings['unifiedCommentsView']"
```

MrMarble (dismissed):
Only a small change that I think is worth doing, the rest looks good!

src/composables/useFlawCommentsModel.ts
```diff
@@ -19,9 +20,24 @@ export function useFlawCommentsModel(flaw: Ref<ZodFlawType>, isSaving: Ref<boole
...
+  );
+
+  const commentsByType = computed<Record<string, ZodFlawCommentType[]>>(() => {
+    return {
+      [CommentType.Public]: publicComments.value.map(c => ({ ...c, type: CommentType.Public })),
+      [CommentType.Private]: privateComments.value.map(c => ({ ...c, type: CommentType.Private })),
+      [CommentType.System]: systemComments.value.map(c => ({ ...c, type: CommentType.System })),
+      [CommentType.Internal]: internalComments.value.map(c => ({ ...c, type: CommentType.Internal })),
+    };
+  });
```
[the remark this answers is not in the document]
C-Valen: The thing is that internal comments doesn't have a computed property, but a ref.
We could use a new computed applying the map to that ref, or do you have any better idea?

src/components/FlawComments/FlawComments.vue
```diff
@@ -1,32 +1,25 @@
...
 import { type ZodFlawCommentType } from '@/types/zodFlaw';
-import { CommentType } from '@/constants';
+import { CommentType, commentTooltips } from '@/constants';
+import { orderCommentsByDate } from '@/utils/helpers';
+import { useSettingsStore } from '@/stores/SettingsStore';
 
 const props = defineProps<{
   bugzillaLink: string;
-  internalComments: ZodFlawCommentType[];
+  commentsByType: Record<string, ZodFlawCommentType[]>;
```
superbuggy, line 19:
```suggestion
  commentsByType: Record<ZodFlawCommentType, ZodFlawCommentType[]>;
```

src/components/FlawComments/FlawComments.vue
```diff
@@ -46,212 +40,90 @@ const commentLabels = computed(() => {
...
+      :bugzillaLink
+      :showJiraLink
+      :isSaving
+      :taskKey
+      :isAddingNewComment
+      :newCommentType
+      :internalCommentsAvailable
+      @startNewComment="(type) => startNewComment(type)"
+    />
+    <div v-if="!internalCommentsAvailable" class="alert alert-danger">
```
MrMarble, line 108: as commented in the code review:
- Add a conditional to only show this alert in the unified view
- Disable the tab and update the tooltip/add an icon to indicate that is not available

MrMarble (approved):

superbuggy (approved):

PR #571, opened 2025-05-20
[OSIDB-3850] Rename incident state

MrMarble:
# [OSIDB-3850] Rename incident state

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Renamed incident states until OSIDB makes the change in OSIDB-3959

## Considerations:

Closes OSIDB-3850


src/types/zodFlaw.ts
```diff
@@ -177,7 +185,7 @@ export const ZodFlawSchema = z.object({
   ),
   meta_attr: z.record(z.string(), z.string().nullish()).nullish(),
   mitigation: z.string().nullish(),
-  major_incident_state: z.nativeEnum(MajorIncidentStateEnumWithBlank).nullish(),
+  major_incident_state: z.nativeEnum(MajorIncidentStateEnumWithBlank).nullable(),
```
superbuggy, line 188: Do you have a preference for nullable over nullish? I think it would make some of the typing simpler to not have to worry about handling undefined and we would 'reclaim' the usefulness of `null`.

C-Valen (approved):

src/types/zodFlaw.ts
MrMarble: From Zod documentation:
>Nullish: To make a schema nullish (both optional and nullable) -> undefined | null
>Nullable: To make a schema nullable (that is, to allow null inputs).

I do not know why, but after my change, typescript was complaining about `undefined` not being a valid value (probably due to this prop being passed through multiple components).

I agree with you about 'reclaiming' null 👍🏼

PR #582, opened 2025-06-23
Update yarn lock file

MrMarble:
#579 removed `axios` but did not update `yarn.lock` file, this throws an error when building the container image as it does not have permissions to write that file


superbuggy:
Just a general note that `axios` here is a secondary/transitive dependency for the OpenAPI generator

superbuggy (approved):

PR #584, opened 2025-06-25
[OSIDB-4271] Cleanup unused/old properties

MrMarble:
# [OSIDB-4271] Cleanup unused/old properties

## Checklist:

- [x] Commits consolidated
- [-] Changelog updated
- [x] Jira ticket updated

## Summary:

There were a lot of unused properties, some of them the property was removed in a refactor but it was still being passed by the parent template, others the property was still set but unused

## Changes:

- Added a couple of ESLint rules to detect unused properties
- enabled `strictTemplates` vue compiler option to throw and error when trying to pass unknown properties to components/elements
- Fixed all the current issues

## Considerations:

Closes OSIDB-4271


MrMarble:
> Nice cleanup.
> 
> How do we confirm that no UI or Functionality were broken with this changes?

Well, I tried my best to test everything and did not find any broken functionality

- Most of the changes are removing unused properties, so that did not affect the UI.
- Other changes, like removing `$attrs`, fixed some bugs, as most of them where used in the root node of the template, which inherits them by default, that's why there are multiple snapshots updated and if you see the diff, is removing duplicated css classes, or invalid properties that where intended to fallthrough to another node.
- Removing `type=button` or `type=text` also does not affect the UI, as those were used in html tags that don't support those attributes anyway.
- `nowrap` is a deprecated html attribute since html 4.1, this could be replaced with css `white-space: nowrap;` but I tested it and there's no change, so I did not implement the css

- Unit tests are passing, only snapshot where updated as I removed attributes, I hope our tests are able of caching those bugs 🤞🏼 

But as with any other pull request, we need to test it thoroughly as we lack proper integration tests right now

src/components/QueryFilterGuide/QueryFilterGuide.vue
```diff
@@ -214,7 +214,7 @@
           <tbody>
             <tr>
               <td>string</td>
-              <td nowrap><code>"this is a string"</code></td>
+              <td><code>"this is a string"</code></td>
```
superbuggy, line 217: Do we want to remove `nowrap` here?

PR #585, opened 2025-06-26
✨ API Keys migration

C-Valen:
# OSIDB-4312 API Keys backend migration

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Instead, OSIM should send a PATCH /osidb/integrations request with the JIRA/BZ token the user has provided.

OSIM should not need to provide the BZ/JIRA HTTP headers when sending requests to OSIDB that require the user's BZ/JIRA tokens, OSIDB will fetch them automatically if the user has already stored them (if not stored, the server will reply with 400 Error so OSIM should still handle the case in which the user has not yet set their tokens).

For direct OSIM-JIRA integration, the user's tokens can be fetched with a GET /osidb/integrations request, the tokens should only be stored in-memory.

The tokens' values should not be displayed back to the user in the settings page.

## Changes:

- **Split settings storage**: API keys now saved securely on backend (/osidb/integrations), UI preferences remain in localStorage
- **New ApiKeyService**: Added `saveApiKeysToBackend()` and `getApiKeysFromBackend()` methods
- **Settings form refactor**: Separate reactive form data for API keys with async backend submission
- **Service updates**: `JiraService` and `OsidbAuthService` updated to fetch API keys from backend instead of localStorage

## Considerations:

- Direct settings store mutation (settingsStore.settings = values) no longer works with split storage (api keys + persistent settings)
- Form inputs needed separate reactive state since API keys now async-loaded from backend
- Headers logic simplified by removing API keys from default OSIDB requests (now fetched per-service)

Closes OSIDB-4312

### DIFF
--- a/src/components/Settings/Settings.vue
+++ b/src/components/Settings/Settings.vue
@@ -1,8 +1,9 @@
 <script setup lang="ts">
-import { ref, watch, reactive } from 'vue';
+import { ref, watch, reactive, computed } from 'vue';
 
 import { useSettingsStore } from '@/stores/SettingsStore';
 import { osimRuntime } from '@/stores/osimRuntime';
+import LoadingSpinner from '@/widgets/LoadingSpinner/LoadingSpinner.vue';
 
 type SensitiveFormInput = 'password' | 'text';
 
@@ -20,6 +21,8 @@ watch(() => settingsStore.apiKeys, (newApiKeys) => {
   formData.bugzillaApiKey = newApiKeys.bugzillaApiKey;
   formData.jiraApiKey = newApiKeys.jiraApiKey;
 }, { immediate: true });
+
+const apiKeysSyncing = computed(() => settingsStore.isLoadingApiKeys || settingsStore.isSavingApiKeys);
 
 const onSubmit = async () => {
   try {
@@ -72,6 +75,7 @@ const isValid = computed(() => ({
             type="radio"
             name="revealSensitive"
             value="password"
+            :disabled="apiKeysSyncing"
           >
           <span class="form-check-label">Hide Password Values</span>
         </label>
@@ -82,6 +86,7 @@ const isValid = computed(() => ({
             type="radio"
             name="revealSensitive"
             value="text"
+            :disabled="apiKeysSyncing"
           >
           <span class="form-check-label">Reveal Password Values</span>
         </label>
@@ -95,6 +100,7 @@ const isValid = computed(() => ({
             class="form-control"
             :type="revealSensitive"
             :class="{'is-invalid': !isValid.bugzillaApiKey,'is-valid': isValid.bugzillaApiKey}"
+            :disabled="apiKeysSyncing"
             placeholder="[none saved]"
           />
           <span
@@ -130,6 +136,7 @@ const isValid = computed(() => ({
             class="form-control"
             :type="revealSensitive"
             :class="{'is-invalid': !isValid.jiraApiKey,'is-valid': isValid.jiraApiKey}"
+            :disabled="apiKeysSyncing"
             placeholder="[none saved]"
           />
           <span
@@ -161,8 +168,14 @@ const isValid = computed(() => ({
         <button
           type="submit"
           class="btn btn-primary"
+          :disabled="apiKeysSyncing"
         >
-          Save Settings
+          <LoadingSpinner
+            v-if="settingsStore.isSavingApiKeys"
+            type="border"
+            class="spinner-border-sm me-2 d-inline-block"
+          />
+          {{ settingsStore.isSavingApiKeys ? 'Saving...' : 'Save Settings' }}
         </button>
       </div>
     </form>
--- a/src/components/Settings/Settings.vue
+++ b/src/components/Settings/Settings.vue
@@ -5,10 +5,10 @@ import { useSettingsStore } from '@/stores/SettingsStore';
 import { osimRuntime } from '@/stores/osimRuntime';
 import LoadingSpinner from '@/widgets/LoadingSpinner/LoadingSpinner.vue';
 
-type SensitiveFormInput = 'password' | 'text';
-
 const settingsStore = useSettingsStore();
-const revealSensitive = ref<SensitiveFormInput>('password');
+
+const showBugzillaKey = ref(false);
+const showJiraKey = ref(false);
 
 // Create reactive form data for API keys only
 const formData = reactive({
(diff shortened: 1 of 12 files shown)


src/stores/SettingsStore.ts
```diff
@@ -31,35 +42,168 @@ const defaultValues: SettingsType = {
...
+        ...newApiKeys,
+      };
+
+      const keysToSave: any = {};
+      if (newApiKeys.bugzillaApiKey !== undefined) {
+        keysToSave.bugzilla = newApiKeys.bugzillaApiKey;
+      }
+      if (newApiKeys.jiraApiKey !== undefined) {
+        keysToSave.jira = newApiKeys.jiraApiKey;
+      }
```
MrMarble, line 182: Couldn't we just use the `newApiKeys`? We are just checking for `undefined` so it's basically copying the keys that exists

src/stores/SettingsStore.ts
```diff
@@ -31,35 +42,168 @@ const defaultValues: SettingsType = {
   unifiedCommentsView: false,
 };
 
+const LEGACY_SETTINGS_KEY = 'OSIM::USER-SETTINGS';
```
MrMarble, line 45: This key is also used for the new config, I don't think it should be named `LEGACY`, also later is used without the constant: https://github.com/RedHatProductSecurity/osim/blob/782c7cffaf07b680db008d05762412b6ef9d7fa8/src/stores/SettingsStore.ts#L113

We could just create a `const SETTINGS_KEY = 'OSIM::USER-SETTINGS';` and use it in the migration function and the rest of the code

src/stores/SettingsStore.ts
```diff
@@ -31,35 +42,168 @@ const defaultValues: SettingsType = {
...
+
+async function migrateApiKeysFromLocalStorage(apiKeys: any, persistentSettings: any, addToast: any) {
```
MrMarble, line 47: This function could be simplified a lot by using the current settings constant:
https://github.com/RedHatProductSecurity/osim/blob/e2d6a1dbc97ae9732aa7b07913cc68fa55a1a0eb/src/stores/SettingsStore.ts#L37
That already handles loading from localStorage and JSON parsing,

```typescript
if (import.meta.env.PROD && (settings.jiraApiKey || settings.bugzillaApiKey)) {
  saveApiKeysToBackend({ jiraApiKey: settings.jiraApiKey, bugzillaApiKey: settings.bugzillaApiKey })
    .then(() => addToast({
      title: 'API Keys Migrated',
      body: 'Your API keys have been securely moved to the server and will no longer be stored in your browser.',
      css: 'success',
    });)
  .catch (() => addToast({
      title: 'Migration Warning',
      body: 'Failed to migrate API keys to server. You may need to re-enter them in Settings.',
      css: 'warning',
    });)
  .finally(() => {
    delete settings.jiraApiKey;
    delete settings.bugzillaApiKey;
  })
}
```

1. Read the current store
2. if there are keys in the store save them to the backend
3. on success notify the user
4. on error notify the user
5. on any case, delete both keys from the store

This will only run once, as after the first run the keys are deleted independently if it fails or not

src/stores/SettingsStore.ts
```diff
@@ -31,35 +42,168 @@ const defaultValues: SettingsType = {
...
+    if (!legacySettingsRaw) {
+      return;
+    }
+
+    let legacySettings;
+    try {
+      legacySettings = JSON.parse(legacySettingsRaw);
+    } catch {
+      return;
+    }
```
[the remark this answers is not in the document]
C-Valen: The thing is that 'getItem' returns `sting | null` and parse requires just `string` so null cases need to be filtered before.

src/stores/SettingsStore.ts
```diff
@@ -31,35 +42,168 @@ const defaultValues: SettingsType = {
...
+      };
+    } catch (error) {
+      console.debug('No API keys found on server or failed to retrieve them');
+    }
   }
 
+  const isTestEnvironment = (typeof window !== 'undefined' && (window as any).__vitest__ !== undefined)
+    || (typeof import.meta !== 'undefined' && import.meta.env?.MODE === 'test');
+
+  if (!isTestEnvironment) {
```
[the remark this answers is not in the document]
C-Valen: Good suggestion

src/stores/SettingsStore.ts
C-Valen: Totally, that'd simplify this

src/components/Settings/Settings.vue
```diff
@@ -1,29 +1,56 @@
...
+  jiraApiKey: Boolean(formData.jiraApiKey),
 });
 
-watch(settings, () => {
+watch(formData, () => {
   isValid.value = {
-    bugzillaApiKey: Boolean(settings.value.bugzillaApiKey),
-    jiraApiKey: Boolean(settings.value.jiraApiKey),
+    bugzillaApiKey: Boolean(formData.bugzillaApiKey),
+    jiraApiKey: Boolean(formData.jiraApiKey),
```
superbuggy, line 53:
```ts
const isValid = computed(() => ({
    bugzillaApiKey: Boolean(formData.bugzillaApiKey),
    jiraApiKey: Boolean(formData.jiraApiKey),
  })
);
```

src/services/ApiKeyService.ts
```diff
@@ -38,6 +52,36 @@ export function notifyApiKeyUnset() {
...
+export async function getApiKeysFromBackend(): Promise<IntegrationTokensResponse> {
+  try {
+    const response = await osidbFetch({
+      method: 'get',
+      url: '/osidb/integrations',
+    });
+    return response.data as IntegrationTokensResponse;
+  } catch (error) {
+    console.error('Failed to retrieve API keys:', error);
+    throw error;
```
superbuggy, line 81: Do we want to re-throw the error or return a value that indicates a fetching failure here?

src/stores/SettingsStore.ts
superbuggy: I wonder if there is a difference between deleting reactive values and setting them to `undefined`, it may be more idiomatic to set them to `undefined`

src/stores/SettingsStore.ts
```diff
@@ -31,35 +42,168 @@ const defaultValues: SettingsType = {
...
+    if (legacySettings.bugzillaApiKey) {
+      keysToMigrate.bugzilla = legacySettings.bugzillaApiKey;
+    }
+    if (legacySettings.jiraApiKey) {
+      keysToMigrate.jira = legacySettings.jiraApiKey;
+    }
+
+    await saveApiKeysToBackend(keysToMigrate);
+
+    const retrievedKeys = await getApiKeysFromBackend();
```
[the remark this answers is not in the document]
superbuggy: If these function calls don't re-throw errors caught in their `catch` statement blocks, I think one level of the try can be eliminated. The function calls can instead return some value related to the error (the error message, `false`/`null`, the error itself).

superbuggy (changes requested):

src/services/ApiKeyService.ts
C-Valen: We don't have a solid pattern for error handling, so I opted for re-throwing the error just cause some operations   legitimately need to fail if they can't get API keys, but not strong preference.

C-Valen:
> I think we could adopt a new error handling pattern where we try to avoid nesting and throwing errors in catch blocks. I have definitely done that in the past and contributed to or started that pattern. We could take a look at that sometime soon if we make time

💯 agreed, I'd love to have some solid and minified error handling for the whole app

PR #377, opened 2024-08-07
Feature/osidb 3129 multi flaws related trackers

superbuggy:
# [OSIDB-ID] [Title]

## Checklist:

- [ ] Commits consolidated
- [ ] Changelog updated
- [ ] Test cases added/updated
- [ ] Jira ticket updated

## Summary:

[Replace with a brief summary of the changes introduced by this PR.]

## Changes:

[Replace with a detailed description of the changes made in this PR, including any new features, enhancements, bug fixes, or refactorings.]

## Considerations:

[Replace with any additional considerations, notes, or instructions for reviewers.]

PR #591, opened 2025-07-10
[OSIDB-4270] Update flaw only when needed

MrMarble:
# [OSIDB-4270] Update flaw only when needed

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

This change reduces the number of network requetsts made when saving changes by detecting if the flaw has been updated or not.
This way if only affects/trackers where modified we skip saving the flaw, reducing network overhead and response times

## Changes:

- Remove `flaw` property from `FlawForm` and `useFlawModel` to use the global `useFlaw` composable
- Removed `refreshAffects` method from `useFlawAffectsModel` since updating flaw/affects already sends a network requests to refresh the data.
- Fixed an issue with the recent implementation of loading affects in parallel, as `initialAffects` was set before `flaw.affects`, breaking almost all interactions with the Affect section
- Removed redundant `type="module"` script tag in `index.html` that was causing some `unexpected MIME type applicaiton/json` errors

## Considerations:

Closes OSIDB-4270


src/composables/useFlaw.ts
```diff
@@ -1,17 +1,29 @@
...
   relatedFlaws.value = [];
 }
 
+function setFlaw(fetchedFlaw: ZodFlawType) {
+  initialFlaw.value = deepCopyFromRaw(fetchedFlaw);
+  flaw.value = fetchedFlaw;
+}
+
+const isFlawUpdated = computed(
+  () => JSON.stringify(flaw.value) !== JSON.stringify(initialFlaw.value),
```
superbuggy, line 22: Changes to affects cause this to return true. If the performance is acceptable, ramda is convenient for this case
```suggestion
  () => R.equals( 
    R.omit(['affects', 'acknowledgments', 'comments', 'references', 'cvss_scores'], flaw.value),
    R.omit(['affects', 'acknowledgments', 'comments', 'references', 'cvss_scores'], initialFlaw),
  )
```

superbuggy: Another approach would be [using a callback with `JSON.stringify`](https://stackoverflow.com/a/46908358) to limit traversal to one level, if that suffices: it would be more performant than using ramda.

src/composables/useFlawAffectsModel.ts
```diff
@@ -39,13 +38,6 @@ export function useFlawAffectsModel() {
     initialAffects.value = deepCopyFromRaw(flaw.value.affects);
   }
 
-  function refreshAffects() {
-    return getFlaw(flaw.value.uuid).then((response) => {
-      flaw.value.affects = [...response.affects];
-      initialAffects.value = [...response.affects];
-    });
-  }
```
superbuggy, line 47: We might need this still until we fully adopt patching Vue state with PUT responses, or with changes to trackers.

superbuggy (changes requested):

src/composables/useFlaw.ts
MrMarble: > Changes to affects cause this to return true. If the performance is acceptable, ramda is convenient for this case
Oh you are right, I must have change something after because I remember trying this and it worked 🤔 

I think the callback approach will be enough

src/composables/useFlawAffectsModel.ts
MrMarble: Oh well you are right again, I need more sleep...

I think I tested this before the change in useFlaw to prevent updating the flaw, saw the duplicated query and removed it, which of course then does not work if the flaw is not updated duh

MrMarble: I was doing some test in stage and this seems to do nothing...

The function is only called in two places:
https://github.com/RedHatProductSecurity/osim/blob/f064cfce07ed95a3b9df446a5eb003858e96e58a/src/composables/useFlawAffectsModel.ts#L263-L265
https://github.com/RedHatProductSecurity/osim/blob/f064cfce07ed95a3b9df446a5eb003858e96e58a/src/components/FlawAffects/FlawAffects.vue#L534-L539

The first one is redundant, since when we update/create an affect, it goes through `useFlawModel` which fetches the flaw/affects after, so this here only duplicates the requests.

The second one is not even used, as there's no `affects:refresh` event, it was removed in this refactor #377

MrMarble: With these change, creating/deleting an affect works as expected. Since we are using the `useFlawModel` to fetch a new flaw/affects, we show the Loading icon for a moment, once we start using the response from creating the affect, this icon should not appear.

Creating trackers also works as expected, when a tracker is created, the `refreshRelatedFlaws` function is called, loading all the affects and thus updating both the trackers and affects table

superbuggy (approved):
looking solid

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## [Unreleased]
 ### Changed
 * Improve loading performance by splitting up flaw and affect requests  (`OSIDB-4266`)
-
+* Don't update Flaw when only affects are modified (`OSIDB-4266`)
```
C-Valen, line 6: Is the ticket correct? PR one points to 4270

C-Valen (dismissed):

PR #598, opened 2025-07-28
✨ Reauthentication banner

C-Valen:
# OSIDB-4275 Reauthentication banner

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Adds a banner that shows up when the user's refresh token is about to expire (less than 1h left).
Supports extending user session (renewing refresh token by reauthenticating) from a button in the warning banner.

## Changes:

- Banner template
- Reauthentication logic
- Make refresh token available in memory
- Update tests

## Considerations:
The `UserStore` changes in this branch are to make the refresh token available in memory to be able to check the expiring time and properly showing the banner.

Closes OSIDB-4275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,15 @@
 # OSIM Changelog
 ## [Unreleased]
+### Added
+* Support extending session (reauthenticating) (`OSIDB-4275`)
+
 ### Changed
 * Move labels table after trackers (`OSIDB-4082`)
 
 ## [2025.8.0]
+### Added
+* Add expiring session banner (`OSIDB-4275`)
+
 ### Fixed
 * Fix only showing 100 affects in flaw view (`OSIDB-4393`)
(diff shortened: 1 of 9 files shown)


superbuggy (commented):
What do you think about putting the auth for users into its own composable/store, and incorporating that within this store? I think the code is generally looking pretty good for this, and I'll do some local testing in the meantime.

C-Valen:
> What do you think about putting the auth for users into its own composable/store, and incorporating that within this store? I think the code is generally looking pretty good for this, and I'll do some local testing in the meantime.

Are you referring to`login/logout/isAuthenticated` form the `userStore`? 
I'd say is a good call to gather all auth-related logic in a composable and make the store use it.

superbuggy (approved):

C-Valen:
Fixed changelog.
Will make a separate PR with the AuthStore refactor.

PR #617, opened 2025-09-03
🚩 Disable CWE AI Suggestions

C-Valen:
Disabling CWE Aegis-AI suggestions for simplifying incoming OSIM releases.

Once Aegis backend is released to prod, we just need to remove the flag and release our side.

## Considerations:
Would be nice to have an actual flag system that allow us to enable/disable features without requiring releases (As @MrMarble suggested a while ago)

### DIFF
--- a/build/entrypoint.d/20-osim-runtime-json.sh
+++ b/build/entrypoint.d/20-osim-runtime-json.sh
@@ -25,6 +25,9 @@ if [ -f "/osim_build.json" ]; then
     IFS= read -r -d '' OSIM_VERSION <"/osim_build.json" || :
 fi
 
+# Feature flags
+OSIM_FLAG_AI_CWE_SUGGESTIONS="${OSIM_FLAG_AI_CWE_SUGGESTIONS:-false}"
+
 IFS= read -r -d '' OSIM_RUNTIME <<EOF || :
 {
   "env": "${OSIM_ENV}",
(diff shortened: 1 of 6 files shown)


MrMarble:
> Would be nice to have an actual flag system that allow us to enable/disable features without requiring releases (As @MrMarble suggested a while ago)

Since we already have the `runtime.json` file that is generated when the container starts, we could modify [20-osim-runtime-json.sh](https://github.com/RedHatProductSecurity/osim/blob/main/build/entrypoint.d/20-osim-runtime-json.sh) to also add any variable that has the prefix `OSIM_FLAG_`.
With that, editing the configmap en OpenShift will expose the variable to the runtime, without any external changes.

So we could discuss this an implement it "easily"

C-Valen:
> > Would be nice to have an actual flag system that allow us to enable/disable features without requiring releases (As @MrMarble suggested a while ago)
> 
> Since we already have the `runtime.json` file that is generated when the container starts, we could modify [20-osim-runtime-json.sh](https://github.com/RedHatProductSecurity/osim/blob/main/build/entrypoint.d/20-osim-runtime-json.sh) to also add any variable that has the prefix `OSIM_FLAG_`. With that, editing the configmap en OpenShift will expose the variable to the runtime, without any external changes.
> 
> So we could discuss this an implement it "easily"

Excellent idea, I have updated this PR to orientate it in that direction.

MrMarble:
https://github.com/RedHatProductSecurity/osim/pull/618

MrMarble (approved):

PR #618, opened 2025-09-03
Automatic flag expansion from environment

MrMarble:
Proof of concept of the possible implementation of feature flag discussed in #617 

This should export every environment variable exposed to the container with the prefix `OSIM_FLAG_`. The prefix is removed, and transformed to camelCase before adding it to `runtime.json`.

For example, given these environment variables:
```env
OSIM_FLAG_CWE_AI_SUGGESTIONS=true
OSIM_FLAG_CVSS_V4=false
OSIM_FLAG_DEFAULT_CVSS_VERSION="V3"
```
Generates this runtime.json:
```jsonc
{
  "featureFlags": {"cvssV4": false,"cweAiSuggestions": true,"defaultCvssVersion": "V3"},
  // other keys in runtime...
}
```

Updating this variables in OpenShift and/or ops repository *and* restarting OSIM (so the `runtime.json` is regenerated) should be the only thing needed when updating the feature flags

### DIFF
--- a/build/entrypoint.d/20-osim-runtime-json.sh
+++ b/build/entrypoint.d/20-osim-runtime-json.sh
@@ -26,8 +26,17 @@ if [ -f "/osim_build.json" ]; then
     IFS= read -r -d '' OSIM_VERSION <"/osim_build.json" || :
 fi
 
-# Feature flags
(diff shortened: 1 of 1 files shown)


C-Valen (approved):
Looks good.
If you have tried locally and works, good to go for me

MrMarble:
I've updated the branch with the latest changes and removed the hardcoded `OSIM_FLAG_AI_CWE_SUGGESTIONS`

PR #640, opened 2025-09-26
[OSIDB-4459] Implements AffectsV2

MrMarble:
# [OSIDB-4459] Implements AffectsV2

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary:

Implemented AffectsV2 using the [TanStack Table](https://tanstack.com/table/latest) library.
This is a full rewrite of the affect section due to the amount of changes and complexity of the current implementation.

With this implementation I tried to separate the logic in layers:

1. The first layer is the data, our affects array, this is unchanged
2. The second layer is `columnDefinitions`, this tells TankStack table **how** to render the data
3. The third layer is `useAffectsModel`, this holds our business logic, application state and methods to integrate with the services
4. The forth and final layer is `AffectsTable`, this integrates the second and third layer to render the table in the UI.

The idea is that if the shape of the data is changed, we only need to update the `columnDefinitions` and maybe a little bit in `useAffectsModel` depending on what has changed. If we need to update how a column is displayed, we only need to change `columnDefinitions` and maybe a bit in `EditableCell` which is the "generic" component  that renders the table cells.
## Changes:

The pull request seems big, but is mostly the update of the generated client from the change in OSIDB api, real changes are around 2.5k lines, including tests and fixture files.

I've separated in commits the setup (install dependencies, configuration changes...), implementation (actual code of the new section with tests) and integration (introduction of the feature flag to use V1 or V2 accordingly), so I hope is not that hard to review.

## Considerations:

Because of the validation with Zod there are some type errors due to the new fields in the AffectsV2 schema.
To fix them I would have to duplicate the `ZodAffect` type and everything that depends on it, since this is only temporal (untile the feature flag is removed) I decided to leave them.

Closes OSIDB-4459


src/services/AffectService.ts
```diff
@@ -118,9 +119,13 @@ export async function putAffectCvssScore(
   delete putObject['uuid'];
   delete putObject['affect'];
   delete putObject['created_dt'];
+
+  const isV2 = osimRuntime.value?.flags?.affectsV2;
+  const apiVersion = isV2 ? 'v2beta' : 'v1';
+  const path = isV2 ? 'cvss-scores' : 'cvss_scores';
```
superbuggy, line 125: :(

src/services/AffectService.ts
```diff
@@ -136,19 +141,25 @@ export async function putAffectCvssScore(
...
     data: postObject,
   })
     .then(response => response.data)
     .catch(createCatchHandler('Problem updating affect CVSS scores:'));
 }
 
 export async function deleteAffectCvssScore(affectId: string, cvssScoresId: string) {
+  const isV2 = osimRuntime.value?.flags?.affectsV2;
+  const apiVersion = isV2 ? 'v2beta' : 'v1';
+  const path = isV2 ? 'cvss-scores' : 'cvss_scores';
```
superbuggy, line 159: wonder if it makes sense to have isV2 or isOsidbApiV2 live in runtime as a readonly

src/components/AffectsTable/customFilters.ts
```diff
@@ -0,0 +1,28 @@
...
+export const arrIncludesWithBlanks: FilterFn<string> = (row, columnId, filterValue: string[]) => {
+  return filterValue.length ? filterValue.includes(row.getValue(columnId)) : true;
+};
+arrIncludesWithBlanks.autoRemove = filterVal => !filterVal?.length;
+
+export const cvssScore: FilterFn<ZodAffectType> = (row, columnId, filterValue: string) => {
+  const rhCvssScore = affectRhCvss3(row.original)?.score || 0;
+  const numberValue = Number.parseFloat(filterValue);
+
+  console.log({ rhCvssScore, numberValue, float: numberValue % 1 !== 0, trunc: Math.trunc(numberValue) });
```
superbuggy, line 22:
```suggestion
```

src/composables/useAffectsModel.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+  }
+
+  async function removeAffects() {
+    await deleteAffects([...removedAffects.values()]);
+    currentAffects.value = currentAffects.value.filter(({ uuid }) => !removedAffects.has(uuid!));
+
+    return [...removedAffects.values()];
+  }
+
```
superbuggy, line 254: Do we essentially only need trackers composables for V1 endpoints?

src/components/AffectsTable/EditableCell.vue
```diff
@@ -0,0 +1,117 @@
...
+</script>
+<template>
+  <span
+    v-if="!editMode"
+    tabindex="0"
+    @dblclick="toggleEditMode"
+    @keyup.tab="toggleEditMode"
+  >{{ displayValue || '&nbsp;' }}</span>
+  <template v-else>
+    <template v-if="metaEnum">
```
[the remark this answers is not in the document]
MrMarble: organizational, since there are is a nested if I though it was easier to read, but can be changed

src/components/AffectsTable/columnDefinitions.tsx
```diff
@@ -0,0 +1,237 @@
...
+        enum: affectImpacts,
+      },
+    }),
+    columnHelper.accessor('cvss_scores', {
+      cell: editableCellRenderer,
+      sortingFn: (rowA, rowB) => {
+        const scoreA = affectRhCvss3(rowA.original)?.score || 0;
+        const scoreB = affectRhCvss3(rowB.original)?.score || 0;
+        const diff = scoreA - scoreB;
+        return diff > 0 ? 1 : diff < 0 ? -1 : 0;
```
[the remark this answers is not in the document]
MrMarble: Oh nice, I wonder how many APIs does Javascript have that I've never used 😆

src/services/AffectService.ts
MrMarble: Yeah... I felt the same. Turns out this change was done a while ago to adhere to http standards (using `-` instead of `_`), but we never used the `v2beta` until now, so it is unrelated to affectsV2 just happens to have the same versioning

src/services/AffectService.ts
```diff
@@ -70,7 +71,7 @@ export async function putAffectWithHandlers(uuid: string, affectObject: any) {
 export async function putAffects(affectObjects: any[]) {
   return osidbFetch({
     method: 'put',
-    url: '/osidb/api/v1/affects/bulk',
+    url: `/osidb/api/${osimRuntime.value?.flags?.affectsV2 ? 'v2beta' : 'v1'}/affects/bulk`,
```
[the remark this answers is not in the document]
C-Valen: Oh yes, forgot that it is temporary. That looks good but keeping it as it is is fine as well.

src/components/AffectsTable/EditableCell.vue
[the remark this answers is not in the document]
superbuggy: Totally agree also makes it clear where it might be broken down further as a component. I like this

src/services/AffectService.ts
superbuggy: my thought was just to move it up so as to only have to change it on one place but probably not worth the effort since it's so little to delete

MrMarble: Ooh you mean outside of the function so it can be reused, yeah 😆  that is a better approach, I though you meant exporting it directly in the `osimRuntime` the `v1`/`v2`.
Do you want me to do the change or is ok?

MrMarble:
@C-Valen Addressed all your comments
> Label column filter does not work, the text for filtering seems not matching the text in the actual labels.

I added the option to search for multiple values using a `,` as a separator
image
> Non affected justification filter is always empty:

Good catch! I actually had that problem for the resolution column and fixed it, but did not notice the justification one
image
>I think column visibility setting should be under the Flaw saving actions div:

I've updated the z-index of the form actions so they appear on top of everything (respecting [bootstrap z-indexes](https://getbootstrap.com/docs/5.0/layout/z-index/#:~:text=We%20don't%20encourage%20customization,%24zindex%2Dtooltip%3A%201080%3B))
image

>Display some text when there are no affects to list

Actually I fixed that before your comment, but it wasn't deployed to UAT 😆 
image

>In some cases "Not affected justification" does not proceed for the affect, being not values in the select. Could we prevent editing the field in those cases?

Done, added a check that if the enum is empty the field is not editable instead of showing that empty selector

C-Valen (commented):
Some further UI/UX considerations:

## Labels Column Filter
I thought the labels were filtered by partial substrings while typing (like other text filters). If we are having exact label/s name matching, and also cause we are supporting multiple values, I'd consider using a dropdown selectable list with the labels to filter (similar to the "non affected justification" filter).
As a "nice to have" thing we could also show the label color in the items of filter.

## Non Affected Justification Column Filter
Could we either have dynamic width in the dropdown element, or at least sightly rising it to avoid line breaks in the current available options:
Image

## Displaying Text For no Affects To List
I was thinking also to have some text for no results with applied filters.

## Empty Select Option
As an Idea, would be nice to rename the empty option in selectable filter list to "_EMPTY". Reasons are 1. It will appear always first in every filter 2. It will become more distinguishable from all other values.

PR #646, opened 2025-10-08
[OSIDB-4524] Tracker suggestions in AffectsV2

MrMarble:
# [OSIDB-4524] Tracker suggestions in AffectsV2

## Checklist:

- [x] Commits consolidated
- [-] Changelog updated
- [x] Jira ticket updated

## Summary:

Tracker suggestions was already a thing in the old affect implementation, due to the Tracker section being removed in the new AffectsV2 the functionality needed to be re implemented.

https://github.com/user-attachments/assets/3caf2a29-dd8c-4481-91dd-b0aaeb1501f9

## Changes:

- Added feature flag usage to tracker service for suggestions endpoint
- Added a new action to the table to query tracker suggestions
- `File tracker` button is replaced with a warning icon if trackers are unavailable
- Suggested trackers are selected and can be created using the bulk action

## Considerations:

Closes OSIDB-4524


superbuggy (approved):
Looks good. Minor comments

C-Valen (commented):
Nice work, I like the toast with the unavailable trackers.

Couple of minor UX considerations:
- IMO there are icons that represent better the functionality, the current one looks more like a "select all rows" action.
e.g.: file-earmark-plus-fill, folder-plus or node-plus-fill

- Should we consider disabling table actions while the bulk tracker filing is being performed? I think currently we allow users change affects meanwhile the bulk operation is in progress.

MrMarble:
> Nice work, I like the toast with the unavailable trackers.

Thanks!

> * IMO there are icons that represent better the functionality, the current one looks more like a "select all rows" action.
>       e.g.: file-earmark-plus-fill, folder-plus or node-plus-fill

Well the functionality does select rows... that's why I went with that icon but I agree is not ideal, I tried your suggestions but doesn't feel right to me neither, what about `list-check` ()  `ui-cheks` () or `ui-checks-grid` ()?


> * Should we consider disabling table actions while the bulk tracker filing is being performed? I think currently we allow users change affects meanwhile the bulk operation is in progress.

Good catch! I'll disable editing the row while the tracker is being filed

C-Valen:
> what about `list-check` () `ui-cheks` () or `ui-checks-grid` ()

From those I like `list-check` but will let it up to you

MrMarble:
@C-Valen I've disabled the editing and row actions when filing trackers, also used a darker shade of gray to highlight that is disabled, I tried with opacity but since the table is gray it didn't look that good

https://github.com/user-attachments/assets/43f12604-db26-43ce-a9d4-9329d981a171

PR #719, opened 2026-01-20
🔧 Fix generated names in Aegis AI client

superbuggy:


C-Valen (commented):
Awesome solution, I didn't know about `modelPropertyNaming`

Just small consideration, should we now remove custom types and use generated client ones on [aegisAI.ts](https://github.com/RedHatProductSecurity/osim/blob/fix-generated-names-in-aegis/src/types/aegisAI.ts)?

superbuggy:
I think so, though not necessarily in this PR. WYDT?

C-Valen (approved):

MrMarble (approved):

PR #762, opened 2026-03-27
👽️ Jira API v3 upgrade

C-Valen:
# OSIDB-4535 Jira API v3 upgrade

## Checklist:

- [x] Commits consolidated
- [-] Changelog updated
- [x] Jira ticket updated

## Summary:

Upgrades OSIM's Jira integration from REST API v2 to v3. Comments are now sent as `Atlassian Document Format (ADF)` and displayed using Jira's server-side pre-rendered HTML.

## Changes:

- All Jira endpoints updated to `/rest/api/3/`
- New `jiraAdf` utility converts Jira markup to ADF for posting comments
- Internal comments display now uses `renderedBody` (pre-rendered HTML from Jira) instead of client-side parsing
- Removed `parseJiraTags` — no longer needed since Jira renders mentions and links server-side

## Considerations:

- Pending Jira stage availability for proper testing
- `linkify` still applied to comments from Bugzilla
- Jira link patterns ([text|url], smart links) were removed from `linkify`
- `?expand=renderedBody` is required on the GET comments request — without it, renderedBody is absent and internal comments will display empty

Closes OSIDB-4545

### DIFF
--- a/src/components/FlawComments/CommentList.vue
+++ b/src/components/FlawComments/CommentList.vue
@@ -65,9 +65,24 @@ async function parseJiraTags(text: string): Promise<string> {
 
 watch(() => props.commentList, async (comments) => {
   isParsingComments.value = true;
-  parsedComments.value = await Promise.all(
-    comments.map(c => parseJiraTags(linkify(sanitizeHtml(c.text ?? '')))),
+
+  // Pre-resolve display names for internal comment authors (accountId → displayName)
+  await Promise.all(
+    comments
+      .filter(c => c.type === CommentType.Internal && c.creator)
+      .map(c => resolveAccountId(c.creator!)),
   );
+
+  parsedComments.value = await Promise.all(
+    comments.map(async (c) => {
+      if (c.type === CommentType.Internal) {
+        // renderedBody is pre-rendered HTML from Jira — only sanitize, skip wiki parsing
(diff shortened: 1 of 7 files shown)


src/components/FlawComments/CommentList.vue
```diff
@@ -100,7 +115,7 @@ function getCommentTooltip(type: CommentType | undefined) {
           :href="jiraUserUrl(comment.creator || '')"
           target="_blank"
         >
-          {{ comment.creator }}
+          {{ userDisplayNameCache[comment.creator ?? ''] ?? comment.creator }}
```
costaconrado: question: is this really valid? I believe whenever comment.creator is null the first part will became `userDisplayNameCache['']` I wonder if this shouldn't be something like
```
comment.creator ? userDisplayNameCache[comment.creator] ?? comment.creator : ''
```

If this lookup for the empty key is intentional feel free to ignore

costaconrado (approved):
I've left a attention point I noticed, please check and if this is the intended behavior, feel free to ignore.

LGTM.

src/components/FlawComments/CommentList.vue
C-Valen: Good catch! Yes that was unintentional, it would have rendered null for empty creators.
Thanks!

PR #774, opened 2026-04-28
♻️ Separate Title and Description suggestions

C-Valen:
♻️ Aegis-293 Separate Title and Description suggestions
## Checklist:

- [x] Commits consolidated
- [x] Changelog updated
- [x] Jira ticket updated

## Summary  
Splits Aegis title and description in indedependent actions so each suggestion applies only its own field.

## Changes  
- New composable for title-only suggestions `AegisTitleActions`.  
- Remove title references from description suggestion component
- Type adjustments
- Tests update

## Considerations  
- Title suggestions still call endpoint **`suggest-description`**  as backend division is not currently on scope

Closes Aegis-293

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # OSIM Changelog
 ## [Unreleased]
+### Added
 * Support lowercase CVE IDs on URLs (Flaw links, advance search and quick search) (`OSIDB-4925`)
+
+### Changed
+* Make Aegis title and description suggestions atomic (`AEGIS-293`)
 
 ## [2026.4.1]
 ### Added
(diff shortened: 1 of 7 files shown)


kdudka (approved):
Works as expected.  Thanks!

MrMarble (approved):
Will the Aegis response be updated to return both suggestions separately? I think that would be a better approach but this LGTM

C-Valen:
> Will the Aegis response be updated to return both suggestions separately? I think that would be a better approach but this LGTM

Good point. We have discussed it but that is more complex than it looks, cause it means changing the actual LLM prompt which implicitly leads to modifying the actual suggestions in both fields.

So not in short-term scope

PR #784, opened 2026-05-18
🐛 Disable reset/save buttons when no flaw changes

C-Valen:
# OSIDB-4973 Disable reset/save buttons when no flaw changes

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated

## Summary

Prevents users from saving or resetting flaws when no changes have been made, avoiding unnecessary API calls and potential programmatic feedback issues.

## Changes

- Disabled "Save Changes" button when no modifications detected
- Disabled "Reset Changes" button when no modifications detected  
- Added tooltips explaining why buttons are disabled
- Detects changes across flaw fields, affects, and CVSS scores

Closes `OSIDB-4973`

### DIFF
--- a/src/components/FlawForm/FlawForm.vue
+++ b/src/components/FlawForm/FlawForm.vue
@@ -102,16 +102,28 @@ const {
   isFetchingAffects,
   totalAffectCount,
 } = useFetchFlaw();
-const { flaw, initialFlaw } = useFlaw();
+const { flaw, initialFlaw, isFlawUpdated } = useFlaw();
 
 const {
   highlightedNvdCvssString,
   nvdCvssString,
   rhCvssString,
   shouldDisplayEmailNistForm,
+  wasFlawCvssModified,
 } = useCvssScores();
 
 const { draftFlaw } = useDraftFlawStore();
+
+const {
+  state: { removedAffects, wereAffectsEditedOrAdded },
+} = useAffectsModel();
+
+const hasAnyChanges = computed(() => {
+  return isFlawUpdated.value
+    || wereAffectsEditedOrAdded.value
+    || wasFlawCvssModified.value
+    || removedAffects.size > 0;
+});
 
 onMounted(() => {
   if (flaw.value.cve_id) {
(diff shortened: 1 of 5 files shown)


src/components/FlawForm/FlawForm.vue
```diff
@@ -103,17 +103,29 @@ const {
...
+const {
+  state: { removedAffects, wereAffectsEditedOrAdded },
+} = useAffectsModel();
+
+const hasAnyChanges = computed(() => {
+  return isFlawUpdated.value
+    || wereAffectsEditedOrAdded.value
+    || wasFlawCvssModified.value
+    || removedAffects.size > 0;
+});
```
MrMarble, line 129: This is missing labels and jira task creation for legacy flaws

```suggestion
const hasAnyChanges = computed(() => {
  return isFlawUpdated.value
    || wereAffectsEditedOrAdded.value
    || wasFlawCvssModified.value
    || removedAffects.size > 0
    || shouldCreateJiraTask.value
    || areLabelsUpdated.value;
});
```

C-Valen: Good catch, and I agree in the labels approach.

For the Jira task I think is better to handle directly in the button condition, so it is not disabled if the checkbox is marked. The reason is cause the Jira task is not something that changed in the flaw but an async operation to be made on the save action.

Does that makes sense to you?

roduran-dev:
I might be wrong here. 

hasAnyChanges now drives button state, but onReset still only restores **flaw**, re inits **affects**, and clears **shouldCreateJiraTask**. 
It does not clear **useFlawLabels**() pending sets (newLabels, updatedLabels, deletedLabels) or reset CVSS via useCvssScores().reset(). After **Reset Changes** 
so  **areLabelsUpdated** / **wasFlawCvssModified** can remain true, so buttons may stay enabled even though the form looks reverted without changes..

C-Valen:
> I might be wrong here.
> 
> hasAnyChanges now drives button state, but onReset still only restores **flaw**, re inits **affects**, and clears **shouldCreateJiraTask**. It does not clear **useFlawLabels**() pending sets (newLabels, updatedLabels, deletedLabels) or reset CVSS via useCvssScores().reset(). After **Reset Changes** so **areLabelsUpdated** / **wasFlawCvssModified** can remain true, so buttons may stay enabled even though the form looks reverted without changes..

@roduran-dev Sorry for late reply. That is a good catch I just pushed a fix for that.
Now `hasAnyChanges` should be correct after resetting changes.

superbuggy (approved):
LGTM if @MrMarble's comment is resolved

roduran-dev (approved):
LGTM

MrMarble (approved):

PR #790, opened 2026-05-27
✨ Flaw state labels system

C-Valen:
# OSIDB-5019 New flaw state label system

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated

## Summary:

Replaces manual state-change buttons with label-driven workflow actions aligned to the new OSIDB workflow framework.

## Changes:

- State field is now read-only; promote/reject/revert/reset buttons removed
- New workflow label type with free text value
- Renamed Contributors section to Labels
- Added `workflow` label type to OpenAPI spec, Zod schema, and regenerated client

## Considerations

Blocks https://github.com/RedHatProductSecurity/osim/pull/803

### DIFF
--- a/src/components/FlawForm/FlawForm.vue
+++ b/src/components/FlawForm/FlawForm.vue
@@ -414,7 +414,7 @@ const isArrayFieldValueAIBot = (fieldName: string, currentValue: null | string[]
         <div class="col-6">
           <FlawWorkflowState
             v-if="mode === 'edit'"
-            :classification="flaw.classification!"
+            :classification="flaw.classification"
             :labels="flaw.labels"
             :shouldCreateJiraTask
             @create:jiraTask="toggleShouldCreateJiraTask()"
--- a/src/components/FlawForm/FlawForm.vue
+++ b/src/components/FlawForm/FlawForm.vue
@@ -415,9 +415,7 @@ const isArrayFieldValueAIBot = (fieldName: string, currentValue: null | string[]
           <FlawWorkflowState
             v-if="mode === 'edit'"
             :classification="flaw.classification"
-            :flawId="flaw.uuid"
             :shouldCreateJiraTask
-            @refresh:flaw="emit('refresh:flaw')"
             @create:jiraTask="toggleShouldCreateJiraTask()"
           />
           <div class="osim-incident-state-container mb-2">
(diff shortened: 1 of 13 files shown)


src/components/FlawForm/FlawForm.vue
```diff
@@ -412,10 +412,9 @@ const isArrayFieldValueAIBot = (fieldName: string, currentValue: null | string[]
         <div class="col-6">
           <FlawWorkflowState
             v-if="mode === 'edit'"
-            :classification="flaw.classification"
-            :flawId="flaw.uuid"
+            :classification="flaw.classification!"
```
MrMarble: I'm not sure about the `!`, what happens with legacy flaws that do not have a workflow classification? if its null this will crash

src/components/FlawLabels/FlawLabelTableEditingRow.vue
```diff
@@ -93,13 +95,13 @@ const emitSave = () => {
     <template v-if="initalLabel?.type === FlawLabelTypeEnum.PRODUCT_FAMILY">
       {{ initalLabel?.label }}
     </template>
-    <!-- Alias labels use free text input -->
-    <template v-else-if="isAliasType">
+    <!-- Alias and workflow labels use free text input -->
+    <template v-else-if="isAliasType || isWorkflowType || initalLabel?.type === FlawLabelTypeEnum.WORKFLOW">
```
MrMarble: Is this condition right? What if the user edits an existing Workflow label? like user edits existing workflow label → changes type dropdown to 'bu' → text input remains instead of BU dropdown because `initalLabel?.type === WORKFLOW` is still true. Is that scenario not possible?

MrMarble (commented):
Small comments to confirm before approving

src/components/FlawForm/FlawForm.vue
C-Valen: adjusted to allow null/undefined in the prop itself

src/components/FlawLabels/FlawLabelTableEditingRow.vue
C-Valen: That's not a possible scenario, label types cannot be changed as they are read only when editing

MrMarble (approved):

PR #802, opened 2026-06-25
🐛 Use email instead username for my issues filter

C-Valen:
# OSIDB-5143 Use email instead username for my issues filter

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated

## Summary:

After Jira migration Owner field must be an email, therefore the my-issues filter needs to work accordingly

## Changes:

- Use email for my issues filter
- Update test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Fixed
 * Disable "Save Changes" and "Reset Changes" buttons when no flaw changes have been made (`OSIDB-4973`)
+* Fix my issues flaw list filter to use email instead username (`OSIDB-5143`)
 
 ### Changed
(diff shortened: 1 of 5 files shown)


MrMarble (approved):
LGTM!

Small question, should we update all jira related fields to use the email? Like `FlawLabelsContributor`, and the Contributors. Maybe already updated, just making sure

superbuggy (approved):

C-Valen:
> LGTM!
> 
> Small question, should we update all jira related fields to use the email? Like `FlawLabelsContributor`, and the Contributors. Maybe already updated, just making sure

@MrMarble Only missing one was selfAssign function from owner field. Updated it to use Jira email, instead username.

PR #809, opened 2026-07-16
🐛 Handle errors in affect operations

C-Valen:
# OSIDB-5196 Handle errors in affect operations

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated

## Summary:

Errors on affect updates were not notified, cause the bulk action response is 200 even if some affects fail, so the failed array in the response was not used.

## Changes:
- `saveAffects`: after each bulk create/update handle failed array to show corresponding notifications
- `resetSavedAffects`: merge saved affects into local state to avoid 409/outdated data errors

### DIFF
--- a/src/composables/useAffectsModel.ts
+++ b/src/composables/useAffectsModel.ts
@@ -2,7 +2,11 @@ import { computed, reactive, readonly, toRaw, shallowRef, triggerRef } from 'vue
 
 import { createSharedComposable } from '@vueuse/core';
 
-import { executeOperationsInParallel, showSuccessToast } from '@/composables/service-helpers';
+import {
+  createCatchHandler,
+  executeOperationsInParallel,
+  showSuccessToast,
+} from '@/composables/service-helpers';
 
 import type { ZodAffectCVSSType, ZodAffectType } from '@/types';
 import {
@@ -46,6 +50,12 @@ function useAffects() {
   }
 
   function resetSavedAffects(savedAffects: ZodAffectType[]) {
+    // Sync server data (e.g. updated_dt) so later edits don't 409
+    if (savedAffects.length) {
+      currentAffects.value = mergeBy(currentAffects.value, savedAffects, 'uuid');
+      initialAffects.value = mergeBy(initialAffects.value, savedAffects, 'uuid');
+    }
+
     // Create a set of saved affect UUIDs for quick lookup
     const savedUuids = new Set(savedAffects.map(affect => affect.uuid));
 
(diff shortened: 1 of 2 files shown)


C-Valen:
@AlejandroMinaya the problem you shared off thread should be fixed now

src/composables/useAffectsModel.ts
```diff
@@ -253,32 +263,48 @@ function useAffects() {
...
+        createCatchHandler(errorTitle, false)(
+          failures.flatMap((f: { errors?: Record<string, string | string[]> }) =>
+            Object.values(f.errors ?? {}).flat(),
+          ).join('\n'),
+        );
+      }
     }
 
-    // Save CVSS scores (this also updates the affects)
     const cvssHasErrors = await saveCvssScores(savedAffects);
```
costaconrado, line 306: I believe we should not call the CVSS score in case the builk affect fails, probably even for newly created affects. (I'm not sure what is the expected behavior here, feel free to ignore in case this is on purpose)

costaconrado (approved):
I left a comment that may not be a real issue, please take a look and feel free to ignore if this is the expected behavior, besides that it LGTM.

src/composables/useAffectsModel.ts
```diff
@@ -253,32 +263,48 @@ function useAffects() {
...
-      ...(toUpdate.length ? [putAffects(toUpdate)] : []),
+    // OSIDB bulk endpoints return 200 with per-item failures in `failed`
+    const bulkOps = [
+      ...(toCreate.length
+        ? [{ errorTitle: 'Error creating Affects:', operation: 'created', request: postAffects(toCreate) }]
+        : []),
+      ...(toUpdate.length
+        ? [{ errorTitle: 'Error updating Affects:', operation: 'updated', request: putAffects(toUpdate) }]
+        : []),
     ];
```
AlejandroMinaya, line 274: __Nitpick__
I believe that since it got more verbose, maybe the ternary operator could become a full-fledged if so it reads better.

src/composables/useAffectsModel.ts
```diff
@@ -253,32 +263,48 @@ function useAffects() {
...
+    for (const [index, { errorTitle, operation }] of bulkOps.entries()) {
+      const settledResult = settled[index];
+      if (settledResult.status === 'rejected' || !settledResult.value?.data) {
+        hasErrors = true;
+        continue;
+      }
+
+      const { data } = settledResult.value;
+      savedAffects.push(...(data.results ?? []));
+      showSuccessToast(data.results?.length ?? 0, 'affect', operation);
```
AlejandroMinaya: **Nitpick**
Since the pushed results and the length reported to the toast depend on the same data, what do you think about setting it as a variable first? Something along these lines:
```suggestion
      const successfulAffects = data.results ?? [];
      savedAffects.push(...successfulAffects);
      showSuccessToast(successfulAffects.length, 'affect', operation);
```

AlejandroMinaya (approved):
Thank you for the fix! I've added two nitpicks but other than that it is good to go! 👍🏾

src/composables/useAffectsModel.ts
C-Valen: Good point, that is a pre-existing pattern. We still needed for saving CVSS on successful affects from `results`.
Skipping when everything fails makes sense for a follow up PR but is not a new scenario introduced from this changes.

Edit: Bug filed as OSIDB-5252

src/composables/useAffectsModel.ts
C-Valen: Fair point, I slightly prefer the compact form here (fewer LOC). In the IDE with highlighting it reads fine to me. GitHub’s diff makes it look denser. Happy to switch if you feel strongly, otherwise I’ll leave it.

src/composables/useAffectsModel.ts
C-Valen: good call, applied!

PR #818, opened 2026-07-24
adding error message for mandatory NOTAFFECTED fields

tgeagle081:
# [OSIDB-ID] [Title]

## Checklist:

- [ ] Commits consolidated
- [ ] Changelog updated

## Summary:

give user an error if trying to save an affect as NOTAFFECTED without providing a "Not Affected Justification". Currently an error comes server side but the user doesn't see it client side leading to confusion. 

## Changes:

File changed: /github/osim/src/types/zodAffect.ts

 Added a .superRefine() cross-field validation to ZodAffectSchema that checks if affectedness is NOTAFFECTED and not_affected_justification is empty/falsy. When  triggered, it adds a Zod issue on the not_affected_justification path with the message: "A justification is required when the affect status is NOTAFFECTED."

## Considerations:

just a suggestion on how to fix it with claude generated code base.

### DIFF
--- a/src/types/zodAffect.ts
+++ b/src/types/zodAffect.ts
@@ -143,6 +143,17 @@ export const ZodAffectSchema = z.object({
   ps_update_stream: z.string(),
   tracker: TrackerSchema.nullable(),
   labels: z.array(z.string()),
+}).superRefine((affect, ctx) => {
+  if (
+    affect.affectedness === AffectednessEnum.Notaffected
+    && !affect.not_affected_justification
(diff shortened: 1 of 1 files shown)


C-Valen:
Hey @tgeagle081 — I think this is the same issue fixed in https://github.com/RedHatProductSecurity/osim/pull/809, which handles error notifications for affect bulk operations. That PR is merged but still pending release, so the fix isn’t in PROD yet.

PR #820, opened 2026-07-31
🐛 Make classification type a string

C-Valen:
# OSIDB-5286 Make classification type a string

## Checklist:

- [x] Commits consolidated
- [-] Changelog updated

## Summary:

Zod validation was rejecting the flaw payload on save because the classification for the flaws is not an anum anymore but a string, making the flaw uneditable for new values not existing in the previous enum.

## Changes:

- Change `classification.state` from enum to string
- Make `workflowState` prop and `postTriageStates` type from enum to string


osoukup (approved):
Looks simple enough to LGTM 😸

PR #834, opened 2026-08-19
💄 Make Affect Module field read-only

superbuggy:
# [OSIDB-4704] Read-only affect module column

## Checklist:

- [x] Commits consolidated
- [x] Changelog updated

## Summary:

Makes ps_module read only, automatically set after save to OSIDB.

## Changes:

Module field in Affects table is now read-only.
Changes to ps_update_stream will cause the previous value of ps_module to appear crossed out.

## Considerations:

[OSIDB-4704]: https://redhat.atlassian.net/browse/OSIDB-4704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # OSIM Changelog
 ## [Unreleased]
 ### Fixed
+* Make `ps_module` field read-only in the affects table; OSIDB determines the module from `ps_update_stream`. Changed values now show a strikethrough until the flaw is saved (`OSIDB-4704`)
 * Allow empty Source Component on flaws in NEW state (`OSIDB-5273`)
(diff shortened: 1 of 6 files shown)


C-Valen (approved):
Looks good.

Found one bug when testing but I think it is pre-existing to this changes:
The CVSS is dropped in new affects creation.

`useAffectsModel` matches the new local row to the OSIDB response on stream + module + component. Local `ps_module` is still empty; OSIDB has already filled it from `ps_update_stream`, so the match fails.

May be fixing it here or in a follow-up: drop `ps_module` from that match and keep `ps_update_stream` + `ps_component`

superbuggy:
Ah yeah, I will fix that bug then

PR #844, opened 2026-09-03
fix: Include the srp_status behind the feature flag

AlejandroMinaya:
## Summary

The `srp_status` field was being included in the `IssueQueueItem` column list unconditionally, leaking the SRP reporting header even when the `srpReporting` feature flag was disabled. This also exposed a pre-existing layout bug where the badge lane's filler `<td>` used `colspan="90%"` — an invalid value that browsers parse as `90`, creating 89 phantom columns and a large blank horizontal space to the right of the table whenever labels were visible.

## Changes

- Converted `nonIdFields` from a static array to a `computed` so it reacts to the feature flag; `srp_status` is now only included when `osimRuntime.value.flags?.srpReporting` is truthy.
- Fixed `colspan="90%"` on the badge lane filler `<td>` to `:colspan="nonIdFields.length"`, so it always spans exactly the remaining columns regardless of the flag state.
- Added a `with srpReporting disabled` test suite covering: header omits SRP Status column, data rows render 6 cells, badge row filler `colspan` is `5`, and labels still render correctly.
- Updated snapshots to reflect the corrected `colspan` value (`6` with SRP on).

### DIFF
--- a/src/components/IssueQueue/IssueQueueItem.vue
+++ b/src/components/IssueQueue/IssueQueueItem.vue
@@ -7,6 +7,7 @@ import SRPStatusBadge from '@/components/CRA/SRPStatusBadge.vue';
 import { useUnprocessedFlawDetection } from '@/composables/unprocessedFlawCheck';
 
 import { labelColorMap } from '@/constants';
+import { osimRuntime } from '@/stores/osimRuntime';
 
 import type { FilteredIssue } from './IssueQueue.vue';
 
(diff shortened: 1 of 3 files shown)


src/components/IssueQueue/IssueQueueItem.vue
```diff
@@ -19,8 +20,14 @@ const { issue, showLabels } = defineProps<{
 // TODO: unhide it once final issue sources are defined. [OSIDB-2424]
 //       and update the CSS for the column width in IssueQueue
 
-const nonIdFields: Exclude<keyof FilteredIssue, 'cve_id' | 'uuid'>[] =
-  ['impact', 'formattedDate', 'title', 'srp_status', 'classification', 'owner'];
+const nonIdFields = computed<Exclude<keyof FilteredIssue, 'cve_id' | 'uuid'>[]>(() => [
+  'impact',
+  'formattedDate',
+  'title',
+  ...(osimRuntime.value.flags?.srpReporting ? ['srp_status' as const] : []),
```
C-Valen, line 27: 👍🏻

C-Valen (approved):
Nice one thanks

PR #836, opened 2026-08-24
🐛 SRP UAT Fixes and enhancements

C-Valen:
# OSIDB-5423 - SRP UAT Fixes and enhancements

## Checklist:

- [x] Commits consolidated
- [ ] Changelog updated
- [x] Test cases added/updated
- [ ] Integration tests updated

## Summary
Integrates changes from UAT feedback, some directly for the frontend and some coming from backend changes

## Changes
- Updated to new report and milestone statuses (OSIDB-5442)
- Added expandable rows, icon sequences, and dynamic ordering
- Added Owner field with self-assign button, Organization Name, and Member States fields
- Made Timer Started At editable with "Set Today" button
- Added visual asterisks to required fields, removed CSIRT fields from payload
- Added support for additional_details JSONField (OSIDB-5482, pending backend merge)
- Fixed flaw comments not appearing until page reload (props to @AlejandroMinaya )

## Considerations:

- Requires backend branch CRA_regulatory_reporting or later (breaking: status changes)
- Requires backend PR for additional details (osidb-5482 in review)
- Deployed in UAT

PR #845, opened 2026-09-04
✨ Expandable milestone rows & fix comment appending

AlejandroMinaya:
## Summary

Two changes for OSIDB-5497:

### ✨ Expandable milestone rows (SRPReportDetails)

`manual_completion_notes` was saved to the backend but never rendered. Adds a per-milestone expand/collapse row (chevron toggle) that reveals:
- Request Source
- Request Received date
- Request Text
- Notes (`manual_completion_notes`)

Keyboard accessible via `Enter`/`Space`. Falls back to *"No additional details."* when all fields are empty.

### 🐛 Fix comment not appended after postFlawComment

After posting a public/private comment, the resolved comment object was passed directly to `afterSaveSuccess()` as its queue argument. `afterSaveSuccess()` expected an array of callbacks and called `.map()` on it, causing a silent TypeError. More critically, `flaw.value.comments` was never updated, so the new comment was invisible until a full page reload.

**Fix:** use the comment object returned by `postFlawComment` to append it to `flaw.value.comments`, then call `afterSaveSuccess()` with no arguments matching its `() => void` signature.

## Tests
- Milestone rows: hidden by default, expand on click showing content, collapse on second click.
- Comment: `should append the saved comment to flaw.value.comments`

### DIFF
--- a/src/components/CRA/SRPReportDetails.vue
+++ b/src/components/CRA/SRPReportDetails.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { ref } from 'vue';
+
 import SRPStatusBadge from '@/components/CRA/SRPStatusBadge.vue';
 
 import type { SRPReport, SRPReportMilestone } from '@/types/cra';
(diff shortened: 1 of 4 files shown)


AlejandroMinaya:
@coderabbitai resolve

C-Valen (commented):
Good work, as we talked off thread the expandable rows it's overlapped from the work on https://github.com/RedHatProductSecurity/osim/pull/836

I took the comment appending fix and added to that PR

PR #849, opened 2026-09-10
feat: Replace additional_details JSON textarea with structured SRP fields

AlejandroMinaya:
## Summary

Replaces the freeform JSON textarea in `SRPMilestoneDialog` with individual form fields based on the SRP reporting table.

### `SRPMilestoneDialog.vue`
- Defines a `SRP_FIELDS` spec (35 fields) covering Common, AEV, and SI sections from the SRP Glossary table
- Fields are filtered by `milestone_type` (24h / 72h / final) and event type scope (AEV / SI); `additional_information_response` milestones hide the section entirely as they have no column in the spec
- Each field displays its requirement status inline: **Required** (red), **Required if available** (amber), **Optional** (muted)
- Required fields block save and show inline `invalid-feedback` messages; `required_if_available` fields are informational only
- Array-valued entries (e.g. `member_states_available`) round-trip correctly: restored as arrays on save if unedited, split on comma if edited
- When `eventType` is unknown, scoped (AEV/SI) fields are hidden to avoid requiring fields from the wrong section

### `SRPSummary.vue`
- Tracks `editingEventType` ref
- Wraps `openEditMilestoneDialog` to capture `report.reportable_event_type` and passes it to the dialog via `:event-type`

### `SRPMilestoneDialog.spec.ts`
- Adds 3 new tests covering: blocked save on empty new milestone, blocked save on empty required SRP fields (72h AEV), and successful save when all required fields are filled

### DIFF
--- a/src/components/CRA/SRPMilestoneDialog.vue
+++ b/src/components/CRA/SRPMilestoneDialog.vue
@@ -3,6 +3,7 @@ import { ref, watch, computed } from 'vue';
 
 import Modal from '@/widgets/Modal/Modal.vue';
 import type {
+  SRPEventType,
   SRPMilestoneStatus,
   SRPMilestoneType,
   SRPReportMilestone,
@@ -10,6 +11,7 @@ import type {
 import { useUserStore } from '@/stores/UserStore';
 
 const props = defineProps<{
+  eventType?: null | SRPEventType;
   milestone?: SRPReportMilestone;
   show: boolean;
 }>();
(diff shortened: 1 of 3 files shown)


AlejandroMinaya:
@coderabbitai review

src/components/CRA/SRPMilestoneDialog.vue
```diff
@@ -21,8 +23,290 @@ const emit = defineEmits<{
...
+  '72h': FieldRequirement;
+  'final': FieldRequirement;
+  'isList'?: true;          // value is a comma-separated list → saved as string[]
+  'key': string;
+  'label': string;
+  'scope': 'aev' | 'common' | 'si';
+  'type': FieldType;
+}
+
+const SRP_FIELDS: SRPFieldSpec[] = [
```
C-Valen, line 44: I think there are keys that don't matck the backend naming:
`manufacturer_or_steward_name`
`general_information`
`general_incident_information`
`aev_detected_at`
`incident_detected_at`

Also I think it worth extract these definitions (also previous interface and types)
in /contstants and /types to reduce the noise in this dialog component

C-Valen (changes requested):
Looks good, great UX enhancement.

Is the backend align with this regarding the validations for each specific report constraints? I think the backend should be the source of truth for this valitdations

Request change is only cause I think the wrong key names in the map might return errors. If it not the case let me know. Thanks!

AlejandroMinaya:
@coderabbitai resolve

C-Valen (approved):
Looks good!

